Issue #29, opened 2015-10-05
Restrict clients on absorber according to their version


abyss7:
Implement some lame check of the client's version to prevent usage of buggy previous versions, or to impose the breakage of backward compatibility.

PR #39, opened 2015-11-09
Add absorber tests

dreamer-dead:
While trying to solve #29 , I decided to add some new Absorber tests.
Also added some helper functions.
Please, take a look.

### DIFF
--- a/src/daemon/absorber_test.cc
+++ b/src/daemon/absorber_test.cc
@@ -6,6 +6,35 @@
 namespace dist_clang {
 namespace daemon {
 
+namespace {
+
+// S1..4 types can be one of the following:
+// |String|, |ConstString|, |Literal| or |const char*|
+template <typename S1, typename S2, typename S3, typename S4 = String>
+net::Connection::ScopedMessage CreateMessage(const S1& source,
+                                             const S2& action,
+                                             const S3& compiler_version,
+                                             const S4& language = S4()) {
+  net::Connection::ScopedMessage message(new net::Connection::Message);
+  auto* extension = message->MutableExtension(proto::Remote::extension);
+  extension->set_source(source);
+  auto* compiler = extension->mutable_flags()->mutable_compiler();
+  compiler->set_version(compiler_version);
+  extension->mutable_flags()->set_action(action);
+  if (language != String())
+    extension->mutable_flags()->set_language(language);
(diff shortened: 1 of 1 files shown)


src/daemon/absorber_test.cc
```diff
@@ -6,6 +6,39 @@
 namespace dist_clang {
 namespace daemon {
 
+namespace {
+
+// S1..4 types can be one of the following:
+// |String|, |ConstString|, |Literal| or |const char*|
```
abyss7: 1. `const char*` is highly discouraged and shouldn't be mentioned.
2. `ConstString` is known as `Immutable`  right now.

src/daemon/absorber_test.cc
```diff
@@ -6,6 +6,39 @@
...
+  auto* compiler = extension->mutable_flags()->mutable_compiler();
+  compiler->set_version(compiler_version);
+  extension->mutable_flags()->set_action(action);
+  if (language != String())
+    extension->mutable_flags()->set_language(language);
+
+  return message;
+}
+
+net::proto::Status StatusFor(const net::proto::Status::Code status_code) {
```
abyss7: No need for `const`

src/daemon/absorber_test.cc
```diff
@@ -239,10 +244,8 @@ TEST_F(AbsorberTest, StoreLocalCache) {
     extension->mutable_flags()->set_action(action);
     extension->mutable_flags()->set_language(language);
```
abyss7: Why not to replace with `CreateMessage()` the code above?

src/daemon/absorber_test.cc
```diff
@@ -6,6 +6,39 @@
...
+                                             const S2& action,
+                                             const S3& compiler_version,
+                                             const S4& language = S4()) {
+  net::Connection::ScopedMessage message(new net::Connection::Message);
+  auto* extension = message->MutableExtension(proto::Remote::extension);
+  extension->set_source(source);
+  auto* compiler = extension->mutable_flags()->mutable_compiler();
+  compiler->set_version(compiler_version);
+  extension->mutable_flags()->set_action(action);
+  if (language != String())
```
abyss7: One-line `if` expression should be inside braces.

PR #40, opened 2015-11-10
Issue 29, check for bad client version

dreamer-dead:
Hi!

Here is my initial version of client version validation (#29), so we can discuss this approach.
Constant 100 for version is a sample only.
Should we expose that minimal version value to C++ code and keep it sync with the default value in proto file?
Please, take a look.

### DIFF
--- a/src/daemon/absorber.cc
+++ b/src/daemon/absorber.cc
@@ -13,6 +13,17 @@ using namespace std::placeholders;
 namespace dist_clang {
 namespace daemon {
 
+namespace {
+
+bool ValidateClientVersion(ui32 client_version) {
+  return client_version >= Absorber::kMinGoodClientVersion;
+}
+
+}  // namespace
+
+// static
+const ui32 Absorber::kMinGoodClientVersion = 100;
+
 Absorber::Absorber(const proto::Configuration& configuration)
     : CompilationDaemon(configuration) {
   using Worker = base::WorkerPool::SimpleWorker;
@@ -67,6 +78,14 @@ bool Absorber::HandleNewMessage(net::ConnectionPtr connection,
 
   if (message->HasExtension(proto::Remote::extension)) {
     Message execute(message->ReleaseExtension(proto::Remote::extension));
+    if (!ValidateClientVersion(execute->client_version())) {
+      LOG(WARNING) << "Client sent a bad version: "
+                   << execute->client_version();
+      net::proto::Status bad_version_status;
+      bad_version_status.set_code(net::proto::Status::BAD_CLIENT_VERSION);
+      return connection->ReportStatus(bad_version_status);
+    }
+
     DCHECK(!execute->flags().compiler().has_path());
     if (execute->has_source()) {
       return tasks_->Push(Task{connection, std::move(execute)});
--- a/src/daemon/absorber_test.cc
+++ b/src/daemon/absorber_test.cc
@@ -10,13 +10,16 @@ namespace {
 
 // S1..4 types can be one of the following: |String|, |Immutable|, |Literal|
 template <typename S1, typename S2, typename S3, typename S4 = String>
-net::Connection::ScopedMessage CreateMessage(const S1& source,
-                                             const S2& action,
-                                             const S3& compiler_version,
-                                             const S4& language = S4()) {
+net::Connection::ScopedMessage CreateMessage(
+    const S1& source,
+    const S2& action,
+    const S3& compiler_version,
+    const S4& language = S4(),
+    ui32 client_version = Absorber::kMinGoodClientVersion) {
   net::Connection::ScopedMessage message(new net::Connection::Message);
   auto* extension = message->MutableExtension(proto::Remote::extension);
   extension->set_source(source);
+  extension->set_client_version(client_version);
   auto* compiler = extension->mutable_flags()->mutable_compiler();
   compiler->set_version(compiler_version);
   extension->mutable_flags()->set_action(action);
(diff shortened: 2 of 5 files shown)


src/daemon/absorber.cc
```diff
@@ -13,6 +13,17 @@ using namespace std::placeholders;
 namespace dist_clang {
 namespace daemon {
 
+namespace {
+
+bool ValidateClientVersion(ui32 client_version) {
```
abyss7: Looking at this function only I can say that this approach is not anticipated. The backward-compatible version should be the field in the Absorber config. Since the usual situation is when the clients (Emitters) are updated much more often than Absorbers, it's convenient to update this field without redistributing the new package.

dreamer-dead: Good point, will do.

dreamer-dead:
Added a new field in Absorber config with tests.
Check for minimum version is quite straightforward, and we can add a max_version to check version in some interval.

src/daemon/remote.proto
```diff
@@ -7,6 +7,7 @@ package dist_clang.daemon.proto;
 message Remote {
   optional base.proto.Flags flags = 1;
   optional string source          = 2;
+  optional uint32 emitter_version = 3 [ default = 100 ];
```
abyss7: Let it be simply "version" - and the default value should be real, i.e. the current version (790.0). Also the real version is complex and consists of two numbers: better to reflect it as a separate message type, like:

    message Version {
      optional uint32 major = 1 [ default = 790 ];
      optional uint32 minor = 2 [ default = 0];
    }

This message type should be shared with a its counterpart in Absorber message.

src/daemon/configuration.proto
```diff
@@ -57,7 +57,8 @@ message Configuration {
   }
 
   message Absorber {
-    required Host local = 1;
+    required Host local                 = 1;
+    optional uint32 min_emitter_version = 2 [ default = 100 ];
```
abyss7: Use shared message type of `version` field from `Remote` message.

src/net/universal.proto
```diff
@@ -9,14 +9,15 @@ message Universal {
...
-    NO_VERSION    = 7;
+    OK                  = 0;
+    INCONSEQUENT        = 1;
+    NETWORK             = 2;
+    BAD_MESSAGE         = 3;
+    EMPTY_MESSAGE       = 4;
+    EXECUTION           = 5;
+    OVERLOAD            = 6;
+    NO_VERSION          = 7;
+    BAD_EMITTER_VERSION = 8;
```
abyss7: Better to choose one or two words, like `BAD_VERSION`, `INCOMPATIBLE`, `DEPRECATED`, etc. and not mention `emitter`.

src/daemon/absorber.cc
```diff
@@ -67,6 +67,15 @@ bool Absorber::HandleNewMessage(net::ConnectionPtr connection,
 
   if (message->HasExtension(proto::Remote::extension)) {
     Message execute(message->ReleaseExtension(proto::Remote::extension));
+    auto config = conf();
+    if (execute->emitter_version() < config->absorber().min_emitter_version()) {
```
abyss7: Implement comparison for the complex version.

src/daemon/absorber.cc
```diff
@@ -67,6 +67,15 @@ bool Absorber::HandleNewMessage(net::ConnectionPtr connection,
...
+      LOG(WARNING) << "Emitter sent a bad version: "
```
abyss7: Better not to mention `emitter` and just say about "incompatible" or "deprecated" client version - not the "bad" one. Also it's better to sync wording with the new name of the Status enum.

src/daemon/absorber_test.cc
```diff
@@ -8,15 +8,24 @@ namespace daemon {
 
 namespace {
 
+static const ui32 kEmptyEmitterVersion = 0u;
```
abyss7: no need for `static` in anonymous namespace.

src/daemon/absorber_test.cc
```diff
@@ -8,15 +8,24 @@ namespace daemon {
...
+static const ui32 kBadEmitterVersion = 1u;
```
abyss7: Anyway, it's better not to use a module-wide constants outside of test bodies - just for couple of them.

abyss7: This constant should be replaced with something like `proto::Remote::default().version() - 1`

src/daemon/absorber_test.cc
```diff
@@ -8,15 +8,24 @@ namespace daemon {
...
+static const ui32 kGoodEmitterVersion = 100u;
```
abyss7: This constant should be replaced with something like `proto::Remote::default().version() + 1`

src/daemon/absorber_test.cc
```diff
@@ -8,15 +8,24 @@ namespace daemon {
...
+net::Connection::ScopedMessage CreateMessage(
+    const S1& source,
+    const S2& action,
+    const S3& compiler_version,
+    const S4& language = S4(),
+    ui32 emitter_version = kEmptyEmitterVersion) {
   net::Connection::ScopedMessage message(new net::Connection::Message);
   auto* extension = message->MutableExtension(proto::Remote::extension);
   extension->set_source(source);
+  if (emitter_version != kEmptyEmitterVersion)
```
abyss7: Forgot braces `{ }`

src/daemon/absorber_test.cc
```diff
@@ -374,5 +383,148 @@ TEST_F(AbsorberTest, BadMessageStatus) {
       << "Daemon must not store references to the connection";
 }
 
+TEST_F(AbsorberTest, BadEmitterVersion) {
```
abyss7: Better not to mention "Emitter".

src/daemon/absorber_test.cc
```diff
@@ -374,5 +383,148 @@ TEST_F(AbsorberTest, BadMessageStatus) {
...
+  EXPECT_EQ(2u, connections_created);
+  EXPECT_EQ(2u, read_count);
+  EXPECT_EQ(2u, send_count);
+  EXPECT_EQ(1, connection1.use_count())
+      << "Daemon must not store references to the connection";
+  EXPECT_EQ(1, connection2.use_count())
+      << "Daemon must not store references to the connection";
+}
+
+TEST_F(AbsorberTest, GoodCustomEmitterVersion) {
```
abyss7: I think, you should modify the existing `SuccessfulCompilation` test instead of making a new one.

src/daemon/absorber_test.cc
```diff
@@ -374,5 +383,148 @@ TEST_F(AbsorberTest, BadMessageStatus) {
...
+    auto message(CreateMessage(""_l, ""_l, ""_l, ""_l, kBadEmitterVersion));
+    EXPECT_TRUE(
+        test_connection->TriggerReadAsync(std::move(message), StatusOK()));
+
+    UniqueLock lock(send_mutex);
+    ASSERT_TRUE(send_condition.wait_for(lock, std::chrono::seconds(1),
+                                        [this] { return send_count == 1; }));
+  }
+
+  auto connection2 = test_service->TriggerListen(expected_host, expected_port);
```
abyss7: This part doesn't correspond for the test's purpose - according to its name. It's better to leave only the first connection.

abyss7:
Done with review.

PR #51, opened 2016-04-15
Added a note about compiler to README, added an option to use gold as a linker

andrey-malets:


build/config/linux/BUILD.gn
```diff
@@ -2,6 +2,7 @@ config("base") {
   defines = [ "OS_LINUX" ]
 
   ldflags = [
+    "-fuse-ld=gold",
```
andrey-malets: This is interpreted as a string actually, but `gold` is treated specially in the clang code: https://github.com/llvm-mirror/clang/blob/master/lib/Driver/Tools.cpp#L10930. Later this string is used to construct linker command line: https://github.com/llvm-mirror/clang/blob/master/lib/Driver/ToolChain.cpp#L358

So I think it's pretty safe to use `gold` here.

PR #52, opened 2016-04-15
A couple of bash code fixes

andrey-malets:


build/run_all_tests
```diff
@@ -1,33 +1,36 @@
 #!/bin/sh
 
-test_dir=`dirname $0`/../out/Test.gn
+set -e
```
andrey-malets: Not sure if this is appropriate, but now the script requires to look at log output from all binaries to ensure everything went OK.

PR #56, opened 2016-06-10
Implement of sending ASAN blacklist to remote compiler

RainM:


src/daemon/absorber.cc
```diff
@@ -137,6 +157,22 @@ void Absorber::DoExecute(const Atomic<bool>& is_shutting_down) {
 
     Universal outgoing(new net::proto::Universal);
 
+    raii_scope_deleter asan_blacklist_deleter;
+    if (incoming->has_asan_blacklist() && incoming->flags().has_asan_blacklist_file()) {
+      g_asan_blacklist_file_write_lock.lock();
```
abyss7: If I understand this code correctly, then you block all absorber threads just to wait for a single temporary file creation. There is a much more simple solution, like using already implemented `base::TemporaryDir` per task execution. You should create this folder at the beginning, put the file with predefined name, e.g. "sanitize_blacklist.txt", and expect that more files may appeat in the flags in the future. All the stuff around this mutex and `TmpUniqFile` seems unnecessary.

src/daemon/compilation_daemon.cc
```diff
@@ -343,6 +343,11 @@ base::ProcessPtr CompilationDaemon::CreateProcess(
   if (flags.has_input()) {
     process->AppendArg(Immutable(flags.input()));
   }
+  if (flags.has_asan_blacklist_file()) {
+    auto flag_value = String("-fsanitize-blacklist="_l);
```
abyss7: If possible please use a couple of tokens, like the flags above, i.e. "-fsanitize-blacklist bla-bla-bla".

src/daemon/emitter.cc
```diff
@@ -389,6 +389,15 @@ void Emitter::DoRemoteExecute(const Atomic<bool>& is_shutting_down,
       continue;
     }
 
+    if (incoming->has_flags() && incoming->flags().has_asan_blacklist_file()) {
+      std::string asan_blacklist_file_path = incoming->current_dir();
+      asan_blacklist_file_path += "//";
```
abyss7: why double-slash?

src/daemon/remote.proto
```diff
@@ -7,6 +7,7 @@ package dist_clang.daemon.proto;
 message Remote {
   optional base.proto.Flags flags = 1;
   optional string source          = 2;
+  optional string asan_blacklist  = 3;
```
abyss7: This flag should be renamed too.

abyss7:
You forgot to consider the blacklist file contents as a part of cache.

abyss7:
Also you should modify existing unit-tests and add new ones.

src/daemon/emitter.cc
RainM: Just to be sure current folder path has slash at the end.

RainM:
I implemented sanitizer blacklist passing to hash. Now, it's possible to add additional 'sources' like sanitizer blacklist to hash.

abyss7:
First of all please run `clang-format` on all modified sources.

src/cache/file_cache.cc
```diff
@@ -114,34 +114,101 @@ bool FileCache::Run(ui64 clean_period) {
...
 
 // static
-HandledHash FileCache::Hash(HandledSource code, CommandLine command_line,
-                            Version version) {
+HandledHash
+FileCache::Hash(
+  HandledSource code,
+  CommandLine command_line,
+  Version version,
+  const std::vector<string::HandledSource>& additional_sources)
```
abyss7: It's better for interface to a single `HandleSource` argument, like: `Vector<HandledSource> sources` - replacing the `code`. The 1-element vector is easy to create with braced-list constructor.

src/cache/file_cache.cc
```diff
@@ -114,34 +114,101 @@ bool FileCache::Run(ui64 clean_period) {
...
+{
+  std::vector<String> hashes;
```
abyss7: Use `Vector` instead of `std::vector` please.

src/cache/file_cache.cc
```diff
@@ -114,34 +114,101 @@ bool FileCache::Run(ui64 clean_period) {
...
+
+  std::transform(
+    std::begin(additional_sources),
+    std::end(additional_sources),
+    std::back_inserter(hashes),
+    [] (const string::HandledSource& hs) -> String {
+      return base::Hexify(hs.str.Hash(4));
```
abyss7: The hash structure should be unified and have a fixed length - it's better to use something like [this](https://github.com/abyss7/dist-clang/blob/fa61909dd2fd8a51d697a0ce4f315c24dd3df798/src/net/test_network_service.h#L9).

src/cache/file_cache.h
```diff
@@ -32,6 +32,7 @@ namespace string {
     explicit type(Immutable str) : str(str) {}                            \
     operator Immutable() const { return str; }                            \
     bool operator==(const type& other) const { return str == other.str; } \
+    bool operator!=(const type& other) const { return str != other.str; } \
```
abyss7: Where is it used?

RainM:
I fixed formatting with clang-format

src/daemon/emitter.cc
```diff
@@ -246,8 +289,19 @@ void Emitter::DoCheckCache(const Atomic<bool>& is_shutting_down) {
       return true;
     };
 
-    if (SearchDirectCache(incoming->flags(), incoming->current_dir(), &entry) &&
-        RestoreFromCache(HandledSource())) {
+    auto additional = ReadUnhandledSources(incoming);
+
+    if (!additional) {
+      LOG(ERROR) << "Can't open additional sources";
```
andrey-malets: What sources? It will be a lot easier to investigate if we put problematic file name and an error which happened while reading it into the message.

src/daemon/emitter.cc
```diff
@@ -389,6 +448,16 @@ void Emitter::DoRemoteExecute(const Atomic<bool>& is_shutting_down,
       continue;
     }
 
+    if (incoming->has_flags() && incoming->flags().has_sanitize_blacklist()) {
+      Immutable content;
+      auto blacklist = incoming->flags().sanitize_blacklist();
+      std::string sanitize_blacklist_file_path = incoming->current_dir();
```
andrey-malets: This code duplicates with the one from `GetAdditionalSourceFiles`, maybe it's better to write a function for it (this could also help moving to a platform where `/` is not a path separator)?

PR #57, opened 2016-06-10
Add arthitecture overview

RainM:


andrey-malets:
Seems that all the terms (dist-clang, emitter, absorber etc) are named with lowercase letters throughout the rest of documentation and code. So maybe it will be better to conform to this rule here too. And if you want to mark out some terms then markdown formatting may be of help.

abyss7:
I'll update the docs anyway - but later. So some description now is better than nothing.

PR #59, opened 2016-06-13
ISSUE-58: use condition variable to notify waiting workers instantly.

andrey-malets:


src/base/worker_pool.h
```diff
@@ -10,18 +10,27 @@ namespace base {
 
 class WorkerPool {
  public:
-  using NetWorker = Fn<void(const Atomic<bool>&, Data&)>;
-  using SimpleWorker = Fn<void(const Atomic<bool>&)>;
+  using NetWorker = Fn<void(WorkerPool*, Data&)>;
```
abyss7: Should be at least `const WorkerPool&` - since pool can't be modified inside worker, and pool can't be `nullptr`

src/base/worker_pool.cc
```diff
@@ -36,11 +38,27 @@ void WorkerPool::AddWorker(Literal name, const NetWorker& worker, ui32 count) {
...
   }
 }
 
+bool WorkerPool::WaitUntilShutdown(
+    const std::chrono::duration<double, std::milli>& duration) {
+  if (is_shutting_down_) {
+    return true;
+  }
+
+  if (duration == std::chrono::duration<double, std::milli>::zero()) {
```
abyss7: This value should be private `static const` member for class - since it's used in more than one place.

src/base/worker_pool.h
```diff
@@ -10,18 +10,27 @@ namespace base {
...
+  using SimpleWorker = Fn<void(WorkerPool*)>;
 
   explicit WorkerPool(bool force_shut_down = false);
   ~WorkerPool();
 
   void AddWorker(Literal name, const NetWorker& worker, ui32 count = 1);
   void AddWorker(Literal name, const SimpleWorker& worker, ui32 count = 1);
 
+  bool WaitUntilShutdown(
+      const std::chrono::duration<double, std::milli>& duration);
```
abyss7: Why `milli`? Why not `seconds`? - You can't specify anywhere millis, i.e. in config.

src/base/worker_pool.cc
```diff
@@ -36,11 +38,27 @@ void WorkerPool::AddWorker(Literal name, const NetWorker& worker, ui32 count) {
...
+    return is_shutting_down_;
+  }
+
+  auto target_time = std::chrono::high_resolution_clock::now() + duration;
+  std::unique_lock<std::mutex> lock(shutdown_mutex_);
+  return shutdown_condition_.wait_until(
```
abyss7: Why not `wait_for`?

src/base/worker_pool.cc
```diff
@@ -36,11 +38,27 @@ void WorkerPool::AddWorker(Literal name, const NetWorker& worker, ui32 count) {
...
+  if (is_shutting_down_) {
+    return true;
+  }
+
+  if (duration == std::chrono::duration<double, std::milli>::zero()) {
+    return is_shutting_down_;
+  }
+
+  auto target_time = std::chrono::high_resolution_clock::now() + duration;
+  std::unique_lock<std::mutex> lock(shutdown_mutex_);
```
abyss7: Should use alias `UniqueLock`

src/base/worker_pool.h
andrey-malets: Made this parameter a reference. Unfortunately, `const` cannot be applied here since worker uses `unique_lock`, which only can get mutex by non-const reference.

src/base/worker_pool.cc
andrey-malets: Done.

src/base/worker_pool.h
andrey-malets: Changed this to seconds.

src/base/worker_pool.cc
andrey-malets: Looked through the doc, `wait_for` seems to be the perfect shortcut :)

src/base/worker_pool.cc
andrey-malets: Done.

src/base/worker_pool.h
abyss7: That's fine: make both mutex and cond. var mutable - they're not part of interface, are they?

andrey-malets: The're rather not, so made them mutable.

src/base/worker_pool_test.cc
```diff
@@ -0,0 +1,35 @@
...
+    UniqueLock lock(mutex);
+    condition.wait(lock, [&ready] { return ready; });
+    EXPECT_TRUE(pool.WaitUntilShutdown(std::chrono::seconds(5)));
+    done = true;
+    condition.notify_one();
+  });
+
+  Thread thread("Test"_l, [&] { pool.reset(); });
+  ready = true;
+  condition.wait_for(lock, std::chrono::seconds(1), [&done] { return done; });
```
abyss7: It would be better to wait a little longer than the pool does - so we can test, that the pool shut down in a specified time frame.

andrey-malets: In this test this durations are expected as we need to check that worker exits early despite trying to wait for a long time. I've added another test below to check the opposite situation.

PR #61, opened 2016-06-16
Enhance thread pool test and make it more strict.

andrey-malets:
Original test probably tests almost nothing: test thread is actually externally synchronised with the thread from thread pool via `ready` and `done` variables: test thread does not continue execution until pooled thread sets `done` to `true` (waiting for one second, which is a huge delay, given that actually it takes less than one millisecond).

For example, we could initialise `WorkerPool` instance with `force_shut_down = true` in `ThreadPool` constructor, which actually brakes the test invariant, and the test would have never noticed that.

New version pushes a bunch of tasks to the queue, so that there would be not enough threads to complete all of them instantly, and then launches the destruction thread, which now really competes with workers. So now we detect broken invariant easily: if we misuse worker pool and force it to shutdown, the queue will have some tasks left and the threads will exit before completing all the tasks.

### DIFF
--- a/src/base/thread_pool_test.cc
+++ b/src/base/thread_pool_test.cc
@@ -17,7 +17,7 @@ TEST(ThreadPoolTest, CompleteAllTasksOnDestruction) {
 
   UniqueLock lock(mutex);
 
-  std::vector<ThreadPool::Optional> futures;
+  Vector<ThreadPool::Optional> futures;
   for (size_t i = 0; i != expected_count; ++i) {
     futures.emplace_back(pool->Push([&] {
       UniqueLock lock(mutex);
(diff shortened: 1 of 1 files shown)


src/base/thread_pool_test.cc
```diff
@@ -9,33 +9,39 @@ TEST(ThreadPoolTest, CompleteAllTasksOnDestruction) {
...
 
   UniqueLock lock(mutex);
 
-  auto future = pool->Push([&] {
-    UniqueLock lock(mutex);
-    condition.wait(lock, [&ready] { return ready; });
-    done = true;
-    condition.notify_all();
-  });
+  std::vector<ThreadPool::Optional> futures;
```
abyss7: `Vector<…>`

andrey-malets: Sure.

PR #62, opened 2016-07-07
Issue 23: implement connection timeout

andrey-malets:


src/net/network_service_test.cc
```diff
@@ -0,0 +1,53 @@
...
+class NetworkServiceTest : public ::testing::Test {
+ public:
+  virtual void SetUp() override {
+    NetworkService::SetFactory<NetworkService::DefaultFactory>();
+  }
+};
+
+TEST_F(NetworkServiceTest, ConnectTimedOut) {
+  auto service = net::NetworkService::Create(0, 0, 1, 1);
+  EndPointPtr unreachable = EndPoint::TcpHost("example.com", 12345, false);
```
andrey-malets: This is pretty silly, I know :) But actually, it seems to do the basic check on a machine with IPv4 connectivity (see https://travis-ci.org/andrey-malets/dist-clang/builds/143170185).

I see three possible solutions probably better than this:
1. Mock entire network stack with a proper library. Requires such a library :)
2. Mangle with `iptables` blocking connections to localhost in the test. Requires super-user privileges to run the test, which I do not find best, plus some OS-specific glue.
3. Disable the test :)

src/client/clang.cc
```diff
@@ -20,13 +20,15 @@ namespace client {
 bool DoMain(int argc, const char* const argv[], Immutable socket_path,
             Immutable clang_path, Immutable version, ui32 read_timeout_secs,
             ui32 send_timeout_secs, ui32 read_min_bytes,
-            const HashMap<String, String>& plugins, bool disabled) {
+            const HashMap<String, String>& plugins, bool disabled,
+            ui32 connect_timeout_secs) {
```
abyss7: It's better to place new argument near similar ones: `read_timeout_secs`, etc.

src/client/clang.cc
```diff
@@ -20,13 +20,15 @@ namespace client {
...
   if (clang_path.empty() || disabled) {
     return true;
   }
 
-  auto service = net::NetworkService::Create(read_timeout_secs,
-                                             send_timeout_secs, read_min_bytes);
+  auto service = net::NetworkService::Create(
+      read_timeout_secs, send_timeout_secs, read_min_bytes,
+      connect_timeout_secs);
```
abyss7: It's better to place new argument before `read_timeout_secs`.

src/client/configuration.proto
```diff
@@ -31,4 +31,7 @@ message Configuration {
   repeated Plugin plugins      = 6;
   optional bool disabled       = 7 [ default = false ];
   optional base.proto.Verbosity verbosity = 8;
+
+  optional uint32 connect_timeout = 9 [ default = 10 ];
```
abyss7: Since the config is not binary encoded, it's better to place new field before `send_timeout`, and even renumber all the fields.

src/net/end_point.h
```diff
@@ -26,6 +30,8 @@ class EndPoint : public std::enable_shared_from_this<EndPoint> {
   int type() const { return SOCK_STREAM; }
   int protocol() const { return protocol_; }
 
+  String Address() const;
```
abyss7: I don't like such interface extension, since the `EndPoint` is abstraction over all possible end-points. Even now unix-socket doesn't have neither host nor port. What do you need it for?

src/client/configuration.proto
```diff
@@ -16,19 +16,22 @@ message Plugin {
...
 
 message Configuration {
-  required string path         = 1;
-  optional string version      = 2;
+  required string path            = 1;
+  optional string version         = 2;
 
-  optional uint32 send_timeout = 3 [ default = 5  ];
+  optional uint32 connect_timeout = 3 [ default = 10 ];
+  // in seconds. Zero means default.
```
abyss7: Do you mean "system default"?

andrey-malets: Actually, when `connect_timeout` is set to zero, we wait as long as connection gets timed out by itself, therefore, yes, this should mean "system default".

andrey-malets: Should I modify a comment to reflect that?

abyss7:
Don't close the task, please.

PR #63, opened 2016-07-11
Remove use of shared_ptr in stat_service.

andrey-malets:


src/perf/stat_service.h
```diff
@@ -18,7 +18,7 @@ class StatService {
   void Dump(proto::Metric& report);
 
  private:
-  Array<SharedPtr<Atomic<ui64>>, proto::Metric::Name_ARRAYSIZE> values_;
```
andrey-malets: Seems to be that `SharedPtr` is no required here.

PR #66, opened 2016-07-22
ISSUE-65: check that input for direct cache is a member of dependencies list.

andrey-malets:


andrey-malets:
Unit tests broke expectedly, so please do not merge it right away.

PR #67, opened 2016-07-22
ISSUE-65: add relative path to source directory as part of the direct cache flags

andrey-malets:


src/daemon/compilation_daemon.cc
```diff
@@ -14,6 +14,17 @@ using namespace cache::string;
 
 namespace {
 
+inline String GetRelativePath(const String current_dir, const String input) {
```
abyss7: Please, describe the semantics of arguments and returned value.

abyss7: Why not `const String&` ?

PR #70, opened 2016-08-01
Support sanitizer blacklist file

Shekeen:
This is a reworked version of https://github.com/abyss7/dist-clang/pull/56

I decided to add a new string type for such files, since they are not actually source files and do not need handling.

I will add tests later.

### DIFF
--- a/src/base/process.h
+++ b/src/base/process.h
@@ -26,6 +26,7 @@ FORWARD_TEST(AbsorberTest, StoreLocalCacheWithAndWithoutBlacklist);
 FORWARD_TEST(CollectorTest, SimpleReport);
 FORWARD_TEST(CompilationDaemonTest, CreateProcessFromFlags);
 FORWARD_TEST(EmitterTest, LocalMessageWithPluginPath);
+FORWARD_TEST(EmitterTest, LocalMessageWithSanitizeBlacklist);
 FORWARD_TEST(EmitterTest, ConfigurationWithoutVersions);
 FORWARD_TEST(EmitterTest, LocalSuccessfulCompilation);
 FORWARD_TEST(EmitterTest, StoreSimpleCacheForLocalResult);
@@ -96,6 +97,7 @@ class Process
   FRIEND_TEST(daemon::CollectorTest, SimpleReport);
   FRIEND_TEST(daemon::CompilationDaemonTest, CreateProcessFromFlags);
   FRIEND_TEST(daemon::EmitterTest, LocalMessageWithPluginPath);
+  FRIEND_TEST(daemon::EmitterTest, LocalMessageWithSanitizeBlacklist);
   FRIEND_TEST(daemon::EmitterTest, ConfigurationWithoutVersions);
   FRIEND_TEST(daemon::EmitterTest, LocalSuccessfulCompilation);
   FRIEND_TEST(daemon::EmitterTest, StoreSimpleCacheForLocalResult);
--- a/src/base/process.h
+++ b/src/base/process.h
@@ -20,7 +20,9 @@ FORWARD_TEST(ClientTest, SendPluginPath);
 }  // namespace client
 
 namespace daemon {
-FORWARD_TEST(AbsorberTest, StoreLocalCache);
+FORWARD_TEST(AbsorberTest, StoreLocalCacheWithoutBlacklist);
+FORWARD_TEST(AbsorberTest, StoreLocalCacheWithBlacklist);
+FORWARD_TEST(AbsorberTest, StoreLocalCacheWithAndWithoutBlacklist);
 FORWARD_TEST(CollectorTest, SimpleReport);
 FORWARD_TEST(CompilationDaemonTest, CreateProcessFromFlags);
 FORWARD_TEST(EmitterTest, LocalMessageWithPluginPath);
@@ -88,7 +90,9 @@ class Process
   FRIEND_TEST(client::ClientTest, SuccessfulCompilation);
   FRIEND_TEST(client::ClientTest, FailedCompilation);
   FRIEND_TEST(client::ClientTest, SendPluginPath);
-  FRIEND_TEST(daemon::AbsorberTest, StoreLocalCache);
+  FRIEND_TEST(daemon::AbsorberTest, StoreLocalCacheWithoutBlacklist);
+  FRIEND_TEST(daemon::AbsorberTest, StoreLocalCacheWithBlacklist);
+  FRIEND_TEST(daemon::AbsorberTest, StoreLocalCacheWithAndWithoutBlacklist);
   FRIEND_TEST(daemon::CollectorTest, SimpleReport);
   FRIEND_TEST(daemon::CompilationDaemonTest, CreateProcessFromFlags);
   FRIEND_TEST(daemon::EmitterTest, LocalMessageWithPluginPath);
--- a/src/base/process.h
+++ b/src/base/process.h
@@ -31,6 +31,7 @@ FORWARD_TEST(EmitterTest, ConfigurationWithoutVersions);
 FORWARD_TEST(EmitterTest, LocalSuccessfulCompilation);
 FORWARD_TEST(EmitterTest, StoreSimpleCacheForLocalResult);
 FORWARD_TEST(EmitterTest, StoreSimpleCacheForRemoteResult);
+FORWARD_TEST(EmitterTest, StoreSimpleCacheForLocalResultWithAndWithoutBlacklist);
 FORWARD_TEST(EmitterTest, StoreDirectCacheForLocalResult);
 FORWARD_TEST(EmitterTest, StoreDirectCacheForRemoteResult);
 FORWARD_TEST(EmitterTest, UpdateConfiguration);
@@ -102,6 +103,7 @@ class Process
   FRIEND_TEST(daemon::EmitterTest, LocalSuccessfulCompilation);
   FRIEND_TEST(daemon::EmitterTest, StoreSimpleCacheForLocalResult);
   FRIEND_TEST(daemon::EmitterTest, StoreSimpleCacheForRemoteResult);
+  FRIEND_TEST(daemon::EmitterTest, StoreSimpleCacheForLocalResultWithAndWithoutBlacklist);
   FRIEND_TEST(daemon::EmitterTest, StoreDirectCacheForLocalResult);
   FRIEND_TEST(daemon::EmitterTest, StoreDirectCacheForRemoteResult);
   FRIEND_TEST(daemon::EmitterTest, UpdateConfiguration);
--- a/src/cache/file_cache.cc
+++ b/src/cache/file_cache.cc
@@ -7,7 +7,6 @@
 #include <perf/stat_service.h>
 
 #include <third_party/snappy/exported/snappy.h>
-#include STL(functional)
 #include STL(regex)
 
 #include <clang/Basic/Version.h>
@@ -28,8 +27,7 @@ String ReplaceTildeInPath(const String& path) {
   return path;
 }
 
-String HashCombine(const Immutable& source,
-                   const HashMap<int, Immutable>& files) {
+String HashCombine(const Immutable& source, const cache::ExtraFiles& files) {
   if (files.empty()) {  // we want hash to be compatible with previous versions
     return base::Hexify(source.Hash());
   }
@@ -129,8 +127,7 @@ bool FileCache::Run(ui64 clean_period) {
 using namespace string;
 
 // static
-HandledHash FileCache::Hash(HandledSource code,
-                            const HashMap<int, Immutable>& extra_files,
+HandledHash FileCache::Hash(HandledSource code, const ExtraFiles& extra_files,
                             CommandLine command_line, Version version) {
   return HandledHash(HashCombine(code.str, extra_files) + "-" +
                      base::Hexify(command_line.str.Hash(4)) + "-" +
@@ -139,7 +136,7 @@ HandledHash FileCache::Hash(HandledSource code,
 
 // static
 UnhandledHash FileCache::Hash(UnhandledSource code,
-                              const HashMap<int, Immutable>& extra_files,
+                              const ExtraFiles& extra_files,
                               CommandLine command_line, Version version) {
   return UnhandledHash(
       HashCombine(code.str, extra_files) + "-" +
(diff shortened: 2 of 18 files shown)


src/cache/file_cache.cc
```diff
@@ -27,6 +27,17 @@ String ReplaceTildeInPath(const String& path) {
...
 }
 
+template <class Container>
+String HashMultipleFiles(const Container& files) {
+  Vector<String> file_hashes;
+  std::transform(std::begin(files), std::end(files),
+                 std::back_inserter(file_hashes),
+                 [](auto&& file) { return base::Hexify(file.str.Hash(4)); });
+  Immutable hashes_string =
+      base::JoinString<'-'>(std::begin(file_hashes), std::end(file_hashes));
```
andrey-malets: https://github.com/abyss7/dist-clang/pull/56#discussion_r68398095

src/cache/file_cache.h
```diff
@@ -101,24 +102,55 @@ class FileCache {
                                   string::CommandLine command_line,
                                   string::Version version);
 
+  static string::HandledHash Hash(string::HandledSource code,
```
andrey-malets: Why duplicate these methods and not modify existing ones? `extra_files` can be empty vector and we can follow existing logic in this case.

Shekeen: Unit tests still use the methods without `extra_files`. I'm not sure if changing all the calls there is better than just having two overloads of these methods.

andrey-malets: I think changing the tests to reflect new signatures should not be hard :) But let's see what @abyss7 says.

abyss7:
There is no need for a new file type: the original idea of `Handled` and `Unhandled` is to overload methods by type. From the type point of view they both are simple strings that are hashed in the same way. I suggest to replace `HandledSource` and `UnhandledSource` with a `List<…>`. If you prefer, you can rename types, i.e. `HandledFile` and `UnhandledFile`.

Shekeen:
If we want to put the contents of (for example) sanitizer blacklist in the `List<T>` together with `HandledSource` or `UnhandledSource`, we need to give these contents the same type. The problems I see here are:

1. This is misleading. These files do not need handling, `HandledSource` and `UnhandledSource` for them are just the same thing, just wrapped into different classes.
2. The only reason to wrap them in `HandledSource` or `UnhandledSource` is to call the correct overload of method (like `FileCache::Find`). But this seems like extra work, why not give these methods different names so that we can call the correct one by specifying its name, not converting the types?

Shekeen:
After looking at the code further I see another problem with keeping blacklist file together with source file, which is related to point №1 in my previous message. If we look at `CompilationDaemon::UpdateDirectCache` method and try to change `const HandledSource& source` to `const List<T>& files` (does not matter what `T` is), we will need to prepare a new list for a `cache_->Store` call, but the first element is the only one that actually has to change. From my point of view we are left with two choices: 

1. Copy all elements but the first into the new list. If we want to use `Handled...` and `Unhandled...` here, we have to convert from `Handled...` to `Unhandled...`, which is counterintuitive. Moreover, this implies that the first element is different from the others, which raises the question: "Why put them in one container?"
2. Read all files from scratch. This is just doing unnecessary work without any justification.

So I think that separating the source file from other auxiliary files is a good idea.

src/daemon/absorber.cc
```diff
@@ -13,6 +13,21 @@ using namespace std::placeholders;
 namespace dist_clang {
 namespace daemon {
 
+namespace {
+
+List<Immutable> GetExtraFiles(proto::Remote* incoming) {
+  List<Immutable> extra_files;
+  std::transform(std::begin(incoming->extra_files()),
```
andrey-malets: Are you sure that this is clearer than the simple loop? :)

```
for (const auto& file : incoming.extra_files())
  extra_files.push_back(ImmutableWrapString(file))
```

src/cache/file_cache.cc
```diff
@@ -27,6 +27,25 @@ String ReplaceTildeInPath(const String& path) {
...
 }
 
+String CombineHashes(const Immutable& first, const Immutable& second) {
+  if (second.empty()) {
+    return first;
+  }
+  return base::Hexify((first + "-"_l + second).Hash());
+}
+
+String HashCombine(const List<Immutable>& files) {
```
andrey-malets:
```
String HashCombine(const List<Immutable>& files) {
  String hashes_string;
  for (auto it = files.begin(); it != files.end(); ++it) {
    if (it != files.begin()) {
      hashes_string += "-";
    }
    hashes_string += base::Hexify(it->Hash());
  }
  return base::Hexify(Immutable(hashes_string).Hash());
}
```

Probably comparing iterator in the input sequence would be easier to inderstand than thinking about empty `hashes_string`.

src/cache/file_cache.cc
```diff
@@ -115,33 +126,60 @@ using namespace string;
 // static
 HandledHash FileCache::Hash(HandledSource code, CommandLine command_line,
                             Version version) {
-  return HandledHash(base::Hexify(code.str.Hash()) + "-" +
+  return FileCache::Hash(code, List<Immutable>{}, command_line, version);
+}
+
+// static
+HandledHash FileCache::Hash(HandledSource code,
```
abyss7: It's better to leave a single `Hash` method: either put the `extra_files` in the end so they can be set to empty default, or simply pass empty brace-initializer everywhere throughout the code.

src/cache/file_cache.cc
```diff
@@ -115,33 +126,60 @@ using namespace string;
...
 }
 
 // static
 UnhandledHash FileCache::Hash(UnhandledSource code, CommandLine command_line,
                               Version version) {
+  return FileCache::Hash(code, List<Immutable>{}, command_line, version);
+}
+
+// static
+UnhandledHash FileCache::Hash(UnhandledSource code,
```
abyss7: It's better to leave a single Hash method: either put the extra_files in the end so they can be set to empty default, or simply pass empty brace-initializer everywhere throughout the code.

src/cache/file_cache.cc
```diff
@@ -115,33 +126,60 @@ using namespace string;
...
           (version.str + "\n"_l + clang::getClangFullVersion()).Hash(4)));
 }
 
 bool FileCache::Find(HandledSource code, CommandLine command_line,
                      Version version, Entry* entry) const {
-  return FindByHash(Hash(code, command_line, version), entry);
+  return Find(code, List<Immutable>{}, command_line, version, entry);
+}
+
+bool FileCache::Find(HandledSource code, const List<Immutable>& extra_files,
```
abyss7: It's better to leave a single `Find` method: either put the extra_files in the end so they can be set to empty default, or simply pass empty brace-initializer everywhere throughout the code.

src/cache/file_cache.cc
```diff
@@ -115,33 +126,60 @@ using namespace string;
...
 }
 
 bool FileCache::Find(UnhandledSource code, CommandLine command_line,
                      Version version, const String& current_dir,
                      Entry* entry) const {
+  return Find(code, List<Immutable>{}, command_line, version, current_dir,
+              entry);
+}
+
+bool FileCache::Find(UnhandledSource code, const List<Immutable>& extra_files,
```
abyss7: It's better to leave a single `Find` method: either put the extra_files in the end so they can be set to empty default, or simply pass empty brace-initializer everywhere throughout the code.

src/cache/file_cache.cc
```diff
@@ -27,6 +27,17 @@ String ReplaceTildeInPath(const String& path) {
   return path;
 }
 
+String HashCombine(const Immutable& source, const List<Immutable>& files) {
+  if (files.empty()) {  // we want hash to be compatible with old versions
+    return source.Hash();
+  }
+  Immutable::Rope hashes_string{source.Hash()};
+  for (auto&& file : files) {
```
abyss7: Why not `const auto&` ? - Since `Hash` method is made `const`.

src/cache/file_cache.cc
```diff
@@ -179,12 +218,27 @@ bool FileCache::Find(UnhandledSource code, CommandLine command_line,
 void FileCache::Store(UnhandledSource code, CommandLine command_line,
                       Version version, const List<String>& headers,
                       const String& current_dir, string::HandledHash hash) {
-  DoStore(Hash(code, command_line, version), headers, current_dir, hash);
+  Store(code, List<Immutable>{}, command_line, version, headers, current_dir,
+        hash);
+}
+
+void FileCache::Store(UnhandledSource code, const List<Immutable>& extra_files,
```
abyss7: It's better to leave a single `Store` method: either put the `extra_files` in the end so they can be set to empty default, or simply pass empty brace-initializer everywhere throughout the code.

src/cache/file_cache.cc
```diff
@@ -179,12 +218,27 @@ bool FileCache::Find(UnhandledSource code, CommandLine command_line,
...
+          hash);
 }
 
 void FileCache::Store(HandledSource code, CommandLine command_line,
                       Version version, const Entry& entry) {
-  DoStore(Hash(code, command_line, version), entry);
+  Store(code, List<Immutable>{}, command_line, version, entry);
+}
+
+void FileCache::Store(HandledSource code, const List<Immutable>& extra_files,
```
abyss7: It's better to leave a single `Store` method: either put the `extra_files` in the end so they can be set to empty default, or simply pass empty brace-initializer everywhere throughout the code.

src/cache/file_cache.h
abyss7: It's better to leave a single method.

src/daemon/absorber.cc
```diff
@@ -13,6 +13,18 @@ using namespace std::placeholders;
 namespace dist_clang {
 namespace daemon {
 
+namespace {
+
+List<Immutable> GetExtraFiles(proto::Remote* incoming) {
```
abyss7: I suggest to directly inline method since it's used only once.

src/daemon/emitter.cc
```diff
@@ -68,6 +68,30 @@ inline bool GenerateSource(const base::proto::Local* WEAK_PTR message,
   return true;
 }
 
+inline bool ReadExtraFiles(const base::proto::Local* WEAK_PTR message,
```
abyss7: Why not use `ReadExtraFiles()` from compilation_daemon.cc ? - I don't see the difference.

abyss7:
I'm waiting for the tests. You can ask me or @andrey-malets for some help. I suggest to fork some existing tests and check at least:
- Cache hit with the same blacklist contents (with different file names).
- Cache miss with different blacklist contents (also check absence of blacklist file).
- Client provides blacklist file name in flags message.

src/cache/file_cache.cc
Shekeen: It's a habit of mine: `auto&&` is shorter to write and the range-based for-loop will use `const_iterator` anyway because of `const List<Immutable>& files`.

Shekeen:
I decided to use `HashMap<int, Immutable>` as a container for extra files. When we use extra files, we need to know which file's contents are we looking at, otherwise we won't know which file it is during remote compilation and we can easily get a false match during hash computation if we use the same contents, but for a different file (not a problem right now, while there is only one possible extra file, but will be a problem if we want to add more extra files).

I also added some tests for `FileCache`, will add more for other components.

Shekeen:
This list (now map) is not just for a sanitizer blacklist file (otherwise why even bother using a container?), but for any possible file that might be used during compilation and passed via command line. There might be several such files, each of them corresponding to a command line argument.

abyss7:
I agree with a `HashMap` - it's a good point. But I'm not sure about the actual implementation: the index of map should have a meaning, i.e. specified by enum'ed values, also it's good to define some type for `HashMap<ExtraType, Immutable>`, like `using ExtrasMap = HashMap<ExtraType, Immutable>` and put it in some relevant class, probably in `FileCache`.

src/cache/file_cache.cc
```diff
@@ -7,6 +7,7 @@
 #include <perf/stat_service.h>
 
 #include <third_party/snappy/exported/snappy.h>
+#include STL(functional)
```
abyss7: Where do you use the functional? - I can't find in the code.

src/cache/file_cache_test.cc
```diff
@@ -155,7 +174,8 @@ TEST(FileCacheTest, RestoreSingleEntry) {
 
   ASSERT_TRUE(base::File::Write(object_path, expected_object_code));
   ASSERT_TRUE(base::File::Write(deps_path, expected_deps));
-  EXPECT_FALSE(cache.Find(code, cl, version, &entry1));
+  EXPECT_FALSE(
+      cache.Find(code, HashMap<int, Immutable>{}, cl, version, &entry1));
```
abyss7: Why to specify the full typename? - At least Clang 3.9.0 should compile fine an unqualified brace-initializer: `cache.Find(code, {}, cl, version, &entry1)` - IMHO, it looks shorter and cleaner.

src/daemon/remote.proto
```diff
@@ -5,8 +5,9 @@ package dist_clang.daemon.proto;
 
 // Sent from emitter to absorber.
 message Remote {
-  optional base.proto.Flags flags = 1;
-  optional string source          = 2;
+  optional base.proto.Flags flags        = 1;
+  optional string source                 = 2;
+  optional string sanitize_blacklist     = 3;
```
abyss7: I don't remember what version of protobuf library right now, but you should try to use the `Map` field (https://developers.google.com/protocol-buffers/docs/proto#maps). If the protobuf is too old, then leave a `FIXME` with a future plans to update protobuf and use fair map here.

src/cache/file_cache_test.cc
Shekeen: It does, I'll change that.

src/cache/file_cache.cc
Shekeen: Oops, I forgot to remove it.

Shekeen:
I already have an index of `HashMap` as a enum, I just didn't specify it explicitly. And moving these declarations to `FileCache` is a good idea, I will do that.

src/daemon/remote.proto
Shekeen: Tried it, received

    daemon/remote.proto:14:3: Expected "required", "optional", or "repeated".
    daemon/remote.proto:14:6: Expected field name.

Seems that protobuf is indeed too old. I will add a `FIXME`

src/base/process.h
```diff
@@ -20,7 +20,9 @@ FORWARD_TEST(ClientTest, SendPluginPath);
 }  // namespace client
 
 namespace daemon {
-FORWARD_TEST(AbsorberTest, StoreLocalCache);
+FORWARD_TEST(AbsorberTest, StoreLocalCacheWoBlacklist);
```
abyss7: What is "Wo" on the test name?

src/cache/file_cache_test.cc
```diff
@@ -15,10 +15,30 @@ using namespace string;
...
+                       hash_regex));
+  EXPECT_TRUE(
+      std::regex_match(FileCache::Hash(HandledSource("1"_l),
+                                       ExtraFiles{{SANITIZE_BLACKLIST, "21"_l}},
+                                       CommandLine("3"_l), Version("4"_l))
+                           .str.string_copy(),
+                       hash_regex));
+}
+
+TEST(FileCacheTest, ExtraFilesGivesDifferentHash) {
```
abyss7: May be `ExtraFilesGiveDifferentHash`?

Shekeen:
The most recent build results: https://travis-ci.org/Shekeen/dist-clang/builds/150906007

PR #71, opened 2016-08-01
Honour possible failure in GetSelfPath.

andrey-malets:


src/client/clang_command.cc
```diff
@@ -120,7 +120,12 @@ bool ClangCommand::FillFlags(base::proto::Flags* flags,
       replaced_command = std::regex_replace(
           replaced_command, version_regex, "/lib/clang/" + clang_major_version);
 
-      const String self_path = base::GetSelfPath();
+      String self_path;
+      String error;
+      if (!base::GetSelfPath(&self_path, &error)) {
+        LOG(WARNING) << "Failed to get executable path: " << error;
```
andrey-malets: In the almost-impossible case when we fail to `GetSelfPath` old version returns empty string, which gets improperly substituted for `clang_path` component later. This version avoids such behaviour and signals to the caller that we were unable to fill flags. This does not seem to be a fatal error, though: the code in `clang.cc` simply executes the command "as is" thereafter.

src/base/c_utils_forward.h
```diff
@@ -13,7 +13,7 @@ Immutable GetCurrentDir(String* error = nullptr);
 Literal GetEnv(Literal env_name, Literal default_env = Literal::empty);
 Literal GetHomeDir(String* error = nullptr);
 void GetLastError(String* error);
-String GetSelfPath(String* error = nullptr);
+bool GetSelfPath(String* result, String* error);
```
abyss7: You've lost the default value for `error` argument.

andrey-malets: Oops :) Fixed.

src/base/c_utils_posix.h
```diff
@@ -75,28 +75,31 @@ inline Literal GetHomeDir(String* error) {
...
     GetLastError(error);
-    return String();
+    return false;
   }
   path[read] = '\0';
 #elif defined(OS_MACOSX)
   ui32 size = sizeof(path);
   if (_NSGetExecutablePath(path, &size) == -1) {
     // TODO: handle not-enough-sized buffer issue.
+    *error = "not enough buffer size";
```
abyss7: Do check `error` for being non-null.

src/base/c_utils_posix.h
```diff
@@ -75,28 +75,31 @@ inline Literal GetHomeDir(String* error) {
...
+    return false;
   }
 // FIXME: convert path to absolute with |realpath()|.
 #else
 #pragma message "Don't know how to get self-path on this platform!"
 #endif
 
-  String&& result = String(path);
-  return result.substr(0, result.find_last_of('/'));
+  *result = String(path);
```
abyss7: Do check `result` for being non-null.

andrey-malets: Added a `CHECK` since we require that argument.

src/base/c_utils_posix.h
andrey-malets: Done.

src/base/c_utils_posix.h
abyss7: If we require it, then it should be a reference.

andrey-malets: OK, changed this to reference.

src/base/c_utils_posix.h
```diff
@@ -75,28 +75,33 @@ inline Literal GetHomeDir(String* error) {
...
   if (read == -1) {
     GetLastError(error);
-    return String();
+    return false;
   }
   path[read] = '\0';
 #elif defined(OS_MACOSX)
   ui32 size = sizeof(path);
   if (_NSGetExecutablePath(path, &size) == -1) {
     // TODO: handle not-enough-sized buffer issue.
```
abyss7: The issue is now handled - remove TODO?

andrey-malets: I thought the real issue is about resizing the buffer (according to the value written to `size` by `_NSGetExecutablePath`) so we can get the path thereafter, and this issue is not resolved yet. Am I wrong?

src/client/configuration.cc
```diff
@@ -104,10 +104,12 @@ Configuration::Configuration() {
     List<String> path_dirs;
     base::SplitString<':'>(path, path_dirs);
 
+    String self_path;
+    CHECK(base::GetSelfPath(self_path, nullptr));
```
abyss7: We should print a warning that we can't reliably detect our-self in `$PATH`. And remove `CHECK`. Also we can abort.

src/client/driver_command.hh
```diff
@@ -28,12 +29,13 @@ class DriverCommand : public Command {
...
+
+  bool CanFillFlags() const override { return !!clang_; }
+
   inline bool FillFlags(base::proto::Flags* flags, const String& clang_path,
                         const String& clang_major_version) const override {
-    if (clang_) {
-      return clang_->FillFlags(flags, clang_path, clang_major_version);
-    }
-    return false;
+    CHECK(clang_);
```
abyss7: It's better to place `DCHECK(CanFillFlags())` - it won't run in release runtime, but will detect logical errors in tests.

src/base/c_utils_posix.h
abyss7: Ok, let's abandon the problem for now. It needs investigation, since I don't know if it possible on Mac to have a path longer than `PATH_MAX`.

PR #74, opened 2016-08-13
Support sanitizer blacklist file

andrey-malets:
* Support sanitize blacklist file

* Fix tests compilation

* Correct hash computation

* Get rid of ExtraFile type, use Immutable instead.

* Remove std::transform and std::accumulate calls.

* Refactor hash computation

* Review fixes

* Use HashMap instead of List for extra files.

* Move ExtraFiles type alias to FileCache, more tests.

* Add more tests

* Remove fsanitize-blacklist flag when preprocessing + one more test.

### DIFF
--- a/src/base/base.proto
+++ b/src/base/base.proto
@@ -43,6 +43,9 @@ message Flags {
   repeated string non_direct   = 10;
   // flags, that shouldn't be neither direct cached, nor cached.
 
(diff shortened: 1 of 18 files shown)


andrey-malets:
The builds: https://travis-ci.org/andrey-malets/dist-clang/builds/152017795

PR #76, opened 2016-09-21
Dump all the values from statistic without arguments.

andrey-malets:


tools/get_report.py
```diff
@@ -50,12 +50,14 @@ def send_message(sock, message):
 
 
 def Main(args):
-    for arg in args[2:]:
+    metrics = args[2:] or [descriptor.name for descriptor in
+                           stat_pb2._METRIC_NAME.values]
```
abyss7: Use enum wrapper interface instead: `stat_pb2.Metric.Name.keys()`

tools/get_report.py
```diff
@@ -50,12 +50,14 @@ def send_message(sock, message):
...
+    for metric in metrics:
         s = socket.create_connection((args[0], args[1]))
 
         m = universal_pb2.Universal()
         r = m.Extensions[stat_pb2.Report.extension].metric.add()
-        r.name = stat_pb2._METRIC_NAME.values_by_name[arg].number
+        r.name = stat_pb2._METRIC_NAME.values_by_name[metric].number
```
abyss7: Same here.

andrey-malets: As `stat_pb2.Metric.Name` is not a `dict`, I used separate `dict` for all key-value pairs.

PR #82, opened 2016-10-07
Fix sanitize blacklist command line argument

Shekeen:
The sanitize blacklist file must be passed via `-fsanitize-blacklist=<file>`, not `-fsanitize-blacklist <file>`.


Shekeen:
Travis builds are green: https://travis-ci.org/Shekeen/dist-clang/builds/165800166

src/base/const_string.h
```diff
@@ -60,6 +60,7 @@ class ConstString {
   ConstString(const Rope& rope);                     // 0-copy
   ConstString(const Rope& rope, size_t hint_size);   // 0-copy
 
+  ConstString(ConstString& str, size_t size);        // 1-copy
```
abyss7: Let's implement the 0-copy variant

Shekeen: Done

PR #90, opened 2016-12-01
Implemented coordinator role.

matthewtff:
Also introduced changes to emitter to make him poll known coordinators
to receive new remote hosts. Also added distribution field for remotes,
but this one is not used for now.

### DIFF
--- a/src/base/empty_lambda.h
+++ b/src/base/empty_lambda.h
@@ -11,7 +11,7 @@ class EmptyLambda {
 
   template <class... Args>
   ReturnType operator()(Args&&...) {
-    return ReturnType();
+    return value_;
   }
 
  private:
--- a/src/base/queue_aggregator.h
+++ b/src/base/queue_aggregator.h
@@ -34,6 +34,7 @@ class QueueAggregator {
     for (auto& thread : threads_) {
       thread.join();
     }
+    threads_.clear();
   }
 
   void Aggregate(LockedQueue<T>* WEAK_PTR queue) THREAD_UNSAFE {
--- a/src/base/queue_aggregator.h
+++ b/src/base/queue_aggregator.h
@@ -34,6 +34,7 @@ class QueueAggregator {
     for (auto& thread : threads_) {
       thread.join();
     }
+
     threads_.clear();
   }
 
--- a/src/daemon/clangd_main.cc
+++ b/src/daemon/clangd_main.cc
@@ -68,7 +68,7 @@ int main(int argc, char* argv[]) {
       coordinator.reset(new daemon::Coordinator(configuration.config()));
     }
 
-    if (coordinator && coordinator->Initialize()) {
+    if (coordinator && !coordinator->Initialize()) {
       LOG(FATAL) << "Coordinator failed to initialize.";
     }
   }
--- a/src/daemon/clangd_main.cc
+++ b/src/daemon/clangd_main.cc
@@ -3,6 +3,7 @@
 #include <daemon/absorber.h>
 #include <daemon/collector.h>
 #include <daemon/configuration.h>
+#include <daemon/coordinator.h>
 #include <daemon/emitter.h>
 
 #include STL(iostream)
@@ -29,7 +30,7 @@ int main(int argc, char* argv[]) {
   signal(SIGSEGV, signal_handler);
 
   daemon::Configuration configuration(argc, argv);
-  UniquePtr<daemon::BaseDaemon> daemon, collector;
+  UniquePtr<daemon::BaseDaemon> daemon, collector, coordinator;
 
   if (configuration.daemonize()) {
 // The function |daemon()| is deprecated on Mac. Use launchd instead.
@@ -50,12 +51,26 @@ int main(int argc, char* argv[]) {
                << configuration.config().user_id();
   }
 
-  if (configuration.config().has_collector()) {
-    collector.reset(new daemon::Collector(configuration.config()));
+  // Initialize collector, if any.
+  {
+    if (configuration.config().has_collector()) {
+      collector.reset(new daemon::Collector(configuration.config()));
+    }
+
+    if (collector && !collector->Initialize()) {
+      LOG(FATAL) << "Collector failed to initialize.";
+    }
   }
 
-  if (collector && !collector->Initialize()) {
-    LOG(FATAL) << "Collector failed to initialize.";
+  // Initialize coordinator, if any.
+  {
+    if (configuration.config().has_coordinator()) {
+      coordinator.reset(new daemon::Coordinator(configuration.config()));
+    }
+
+    if (coordinator && coordinator->Initialize()) {
+      LOG(FATAL) << "Coordinator failed to initialize.";
+    }
   }
 
   if (configuration.config().has_absorber()) {
--- a/src/daemon/collector.cc
+++ b/src/daemon/collector.cc
@@ -13,13 +13,11 @@ namespace daemon {
 Collector::Collector(const proto::Configuration& configuration)
     : BaseDaemon(configuration) {
   CHECK(configuration.has_collector());
-  conf_.reset(new proto::Configuration(configuration));
-  conf_->CheckInitialized();
 }
 
 bool Collector::Initialize() {
   String error;
-  const auto& local = conf_->collector().local();
+  const auto& local = conf()->collector().local();
   if (!Listen(local.host(), local.port(), local.ipv6(), &error)) {
     LOG(ERROR) << "[Collector] Failed to listen on " << local.host() << ":"
                << local.port() << " : " << error;
--- a/src/daemon/collector.cc
+++ b/src/daemon/collector.cc
@@ -11,15 +11,18 @@ namespace dist_clang {
 namespace daemon {
 
 Collector::Collector(const proto::Configuration& configuration)
-    : BaseDaemon(configuration), local_(configuration.collector().local()) {
+    : BaseDaemon(configuration) {
   CHECK(configuration.has_collector());
+  conf_.reset(new proto::Configuration(configuration));
+  conf_->CheckInitialized();
 }
 
 bool Collector::Initialize() {
   String error;
-  if (!Listen(local_.host(), local_.port(), local_.ipv6(), &error)) {
-    LOG(ERROR) << "Failed to listen on " << local_.host() << ":"
-               << local_.port() << " : " << error;
+  const auto& local = conf_->collector().local();
+  if (!Listen(local.host(), local.port(), local.ipv6(), &error)) {
+    LOG(ERROR) << "[Collector] Failed to listen on " << local.host() << ":"
+               << local.port() << " : " << error;
     return false;
   }
 
--- a/src/daemon/collector.cc
+++ b/src/daemon/collector.cc
@@ -12,14 +12,14 @@ namespace daemon {
 
 Collector::Collector(const proto::Configuration& configuration)
     : BaseDaemon(configuration) {
-  CHECK(configuration.has_collector());
+  CHECK(conf()->has_collector());
 }
 
 bool Collector::Initialize() {
   String error;
   const auto& local = conf()->collector().local();
   if (!Listen(local.host(), local.port(), local.ipv6(), &error)) {
-    LOG(ERROR) << "[Collector] Failed to listen on " << local.host() << ":"
+    LOG(ERROR) << "Collector failed to listen on " << local.host() << ":"
                << local.port() << " : " << error;
     return false;
   }
--- a/src/daemon/common_daemon_test.h
+++ b/src/daemon/common_daemon_test.h
@@ -15,6 +15,7 @@ namespace daemon {
 
 struct CommonDaemonTest : public ::testing::Test {
   using Service = net::TestNetworkService;
+  using ServiceCallback = Fn<void(Service*)>;
   using ListenCallback = Fn<bool(const String&, ui16, String*)>;
   // Reject connection by returning |false| from |connect_callback|.
   using ConnectCallback = Fn<bool(net::TestConnection*, net::EndPointPtr)>;
@@ -28,6 +29,7 @@ struct CommonDaemonTest : public ::testing::Test {
         test_service = service;
         service->CountConnectAttempts(&connect_count);
         service->CountListenAttempts(&listen_count);
+        service_callback(service);
         service->CallOnConnect([this](net::EndPointPtr end_point, String*) {
           auto connection = net::TestConnectionPtr(new net::TestConnection);
           connection->CountSendAttempts(&send_count);
@@ -71,6 +73,7 @@ struct CommonDaemonTest : public ::testing::Test {
     net::EndPointResolver::SetFactory<net::TestEndPointResolver::Factory>();
   }
 
+  ServiceCallback service_callback = EmptyLambda<>();
   ListenCallback listen_callback = EmptyLambda<bool>(true);
   ConnectCallback connect_callback = EmptyLambda<bool>(true);
   RunCallback run_callback = EmptyLambda<>();
--- a/src/daemon/common_daemon_test.h
+++ b/src/daemon/common_daemon_test.h
@@ -17,7 +17,6 @@ struct CommonDaemonTest : public ::testing::Test {
   using Service = net::TestNetworkService;
   using ServiceCallback = Fn<void(Service*)>;
   using ListenCallback = Fn<bool(const String&, ui16, String*)>;
-  // Reject connection by returning |false| from |connect_callback|.
   using ConnectCallback = Fn<bool(net::TestConnection*, net::EndPointPtr)>;
   using RunCallback = Fn<void(base::TestProcess*)>;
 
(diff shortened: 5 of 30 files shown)


src/daemon/configuration.proto
```diff
@@ -1,13 +1,17 @@
 import "base/base.proto";
+import "net/universal.proto";
 
 package dist_clang.daemon.proto;
 
 message Host {
   required string host          = 1;
   optional uint32 port          = 2 [ default = 6000 ];
+  // Ignored for coordinators and collectors.
```
abyss7: Put the comments below the commented line and insert empty lines for better visualization.

src/daemon/configuration.proto
```diff
@@ -35,10 +39,15 @@ message Configuration {
...
-    required string socket_path = 1;
-    repeated Host remotes       = 2;
-    optional uint32 threads     = 3 [ default = 2 ];
-    optional bool only_failed   = 4 [ default = false ];
+    required string socket_path                = 1;
+    repeated Host remotes                      = 2;
+    optional uint32 threads                    = 3 [ default = 2 ];
+    optional bool only_failed                  = 4 [ default = false ];
+
+    optional uint32 poll_coordinators_interval = 5 [ default = 30 ];
```
abyss7: It's better to shorten the name (i.e. `poll_interval`) and add proper comment about coordinators.

src/daemon/configuration.proto
```diff
@@ -52,33 +61,44 @@ message Configuration {
     required Host local = 1;
   }
 
+  message Coordinator {
+    required Host local   = 1;
+    repeated Host remotes = 2;
```
abyss7: If this field is about remotes to propagate, then add comment about it, please.

src/daemon/coordinator.h
```diff
@@ -0,0 +1,27 @@
...
+
+namespace dist_clang {
+namespace daemon {
+
+class Coordinator : public BaseDaemon {
+ public:
+  explicit Coordinator(const proto::Configuration& configuration);
+
+  bool Initialize() override;
+  inline bool UpdateConfiguration(
```
abyss7: Why to override with the same body?

src/daemon/coordinator.h
```diff
@@ -0,0 +1,27 @@
...
+      const proto::Configuration& configuration) override {
+    return true;
+  }
+
+ private:
+  bool HandleNewMessage(net::ConnectionPtr connection, Universal message,
+                        const net::proto::Status& status) override;
+
+  const proto::Host local_;
```
abyss7: Why not to use `SharedPtr<const proto::Configuration> conf_;` like in `CompilationDaemon`? - It's potentially easier this way to reload config from local file system. Also, you preserve the config structure only in one place - the `.proto` file.

src/daemon/coordinator.cc
```diff
@@ -0,0 +1,64 @@
...
+      local_(configuration.coordinator().local()),
+      remotes_(configuration.coordinator().remotes().begin(),
+               configuration.coordinator().remotes().end()) {
+  CHECK(configuration.has_coordinator());
+}
+
+bool Coordinator::Initialize() {
+  String error;
+  if (!Listen(local_.host(), local_.port(), local_.ipv6(), &error)) {
+    LOG(ERROR) << "[Coordinator] Failed to listen on " << local_.host() << ":"
```
abyss7: Apply this message format to all other daemons.

src/daemon/coordinator.cc
```diff
@@ -0,0 +1,64 @@
...
+  if (status.code() != net::proto::Status::OK) {
+    LOG(ERROR) << status.description();
+    return connection->ReportStatus(status);
+  }
+
+  if (message->HasExtension(proto::Configuration::extension)) {
+    UniquePtr<proto::Configuration> configuration(
+        message->ReleaseExtension(proto::Configuration::extension));
+
+    proto::Configuration::Emitter* emitter = configuration->mutable_emitter();
```
abyss7: You should explicitly clear all other fields - to prevent possible misconfiguration on the other side.

src/daemon/coordinator.cc
```diff
@@ -0,0 +1,64 @@
...
+    emitter->set_socket_path("no-op");
```
abyss7: Please, add FIXME about migration to Protobuf 3, where there is no `required` fields. This line will be unnecessary.

src/daemon/coordinator.cc
```diff
@@ -0,0 +1,64 @@
...
+    for (const proto::Host& remote : remotes_) {
+      proto::Host* host = emitter->add_remotes();
+      *host = remote;
```
abyss7: It's better to explicitly call `CopyFrom()`.

src/base/queue_aggregator.h
```diff
@@ -34,6 +34,7 @@ class QueueAggregator {
     for (auto& thread : threads_) {
       thread.join();
     }
+    threads_.clear();
```
abyss7: What is it for?

src/daemon/coordinator.h
matthewtff: Looks like it's a debt. I've tried to use `UpdateConfiguration` from `DoCoordinator`, but that one requires fully specified `Configuration` message, which is not a case for coordinator. I'll remove it.

src/base/queue_aggregator.h
matthewtff: To be able to call it twice for `QueueAggregator` - used in tests. Otherwise it would try
to join already joined threads.

src/daemon/coordinator.cc
abyss7: Looks like you simply need `configuration->Clear()`

src/daemon/emitter.cc
```diff
@@ -527,5 +582,63 @@ void Emitter::DoRemoteExecute(const base::WorkerPool& pool,
   }
 }
 
+void Emitter::DoCoordinator(const base::WorkerPool& pool,
```
abyss7: Let's name the method "do _verb_", like `DoPoll`.

src/daemon/emitter.cc
```diff
@@ -527,5 +582,63 @@ void Emitter::DoRemoteExecute(const base::WorkerPool& pool,
...
+                            std::vector<ResolveFn> resolvers) {
```
abyss7: `Vector`

src/daemon/emitter.cc
```diff
@@ -527,5 +582,63 @@ void Emitter::DoRemoteExecute(const base::WorkerPool& pool,
...
+              << " seconds before next attempt";
+    std::this_thread::sleep_for(std::chrono::seconds(sleep_period));
+    if (sleep_period < static_cast<ui32>(-1) / 2) {
+      sleep_period <<= 1;
+    }
+  };
+
+  compilation_workers_.reset(new base::WorkerPool(true));
+  while (!pool.IsShuttingDown()) {
+    if (!connection) {
```
abyss7: It's not meant by design to keep connection alive.

src/daemon/emitter.cc
```diff
@@ -527,5 +582,63 @@ void Emitter::DoRemoteExecute(const base::WorkerPool& pool,
...
+      for (const ResolveFn& resolver : resolvers) {
```
abyss7: I don't like the concept of round-robin the coordinators in a single thread - at least you don't have a separate sleep time for each one. Let's think a better way. May be setup a worker per coordinator - with a common atomic variable about last update. Run them with a time shift about 5 minutes. If the first one doesn't update config before the second one wakes up, then the second one tries to update config, and so on.

src/daemon/emitter.cc
```diff
@@ -527,5 +582,63 @@ void Emitter::DoRemoteExecute(const base::WorkerPool& pool,
...
+        end_point = resolver();
+        if (!end_point)
+          continue;
```
abyss7: One-liners without braces are forbidden by code style.

src/daemon/emitter.cc
```diff
@@ -527,5 +582,63 @@ void Emitter::DoRemoteExecute(const base::WorkerPool& pool,
...
+      }
+    }
+    if (!connection) {
+      Sleep();
+      continue;
+    }
+
+    sleep_period = 1;
+    UniquePtr<proto::Configuration> configuration(new proto::Configuration);
+    if (!connection->SendSync(std::move(configuration)))
```
abyss7: Anyway you should iterate the full send-read cycle per coordinator - in case you can connect, but doesn't get a response.

src/daemon/emitter.cc
```diff
@@ -527,5 +582,63 @@ void Emitter::DoRemoteExecute(const base::WorkerPool& pool,
...
+      continue;
+
+    Universal reply(new net::proto::Universal);
+    if (!connection->ReadSync(reply.get()))
+      continue;
+
+    if (reply->HasExtension(proto::Configuration::extension)) {
+      auto* config = reply->MutableExtension(proto::Configuration::extension);
+      UpdateCompilationPool(*config);
```
abyss7: There is already framework for a configuration update mechanics: look at https://github.com/abyss7/dist-clang/blob/master/src/daemon/compilation_daemon.h#L13 - why to use a separate method?

src/daemon/emitter.h
```diff
@@ -36,13 +38,22 @@ class Emitter : public CompilationDaemon {
...
   void DoCheckCache(const base::WorkerPool&);
   void DoLocalExecute(const base::WorkerPool&);
   void DoRemoteExecute(const base::WorkerPool&, ResolveFn resolver);
+  void DoCoordinator(const base::WorkerPool&, std::vector<ResolveFn> resolvers);
 
   UniquePtr<Queue> all_tasks_, cache_tasks_, failed_tasks_;
   UniquePtr<QueueAggregator> local_tasks_;
   UniquePtr<base::WorkerPool> workers_;
+  UniquePtr<base::WorkerPool> compilation_workers_;
+  Vector<proto::Host> remotes_;
```
abyss7: This should be an internal part of `conf_` - https://github.com/abyss7/dist-clang/blob/master/src/daemon/compilation_daemon.h#L59
See comment about `UpdateConfiguration()`.

src/daemon/emitter.h
```diff
@@ -36,13 +38,22 @@ class Emitter : public CompilationDaemon {
...
+
   void DoCheckCache(const base::WorkerPool&);
   void DoLocalExecute(const base::WorkerPool&);
   void DoRemoteExecute(const base::WorkerPool&, ResolveFn resolver);
+  void DoCoordinator(const base::WorkerPool&, std::vector<ResolveFn> resolvers);
 
   UniquePtr<Queue> all_tasks_, cache_tasks_, failed_tasks_;
   UniquePtr<QueueAggregator> local_tasks_;
   UniquePtr<base::WorkerPool> workers_;
+  UniquePtr<base::WorkerPool> compilation_workers_;
```
abyss7: It's better to name like `remote_workers_`

abyss7: Why not a `SharedPtr` as we discussed? Either you will hang on `reset()` or lose all current jobs.

src/daemon/emitter.cc
matthewtff: I tried to use that method, but it resulted as overcomplicated code. It's called from `Initialize` - what for? So it could be called on different threads. As a result more difficult pool terminating mechanics to make sure only thread safe methods are used.

src/daemon/emitter.h
matthewtff: Looks like it's ok to hang on reset() once new pool already created, anyway `sleep()` goes next.

src/daemon/collector.cc
```diff
@@ -11,15 +11,18 @@ namespace dist_clang {
 namespace daemon {
 
 Collector::Collector(const proto::Configuration& configuration)
-    : BaseDaemon(configuration), local_(configuration.collector().local()) {
+    : BaseDaemon(configuration) {
   CHECK(configuration.has_collector());
+  conf_.reset(new proto::Configuration(configuration));
```
abyss7: Since all descendants of `BaseDaemon` now have the `conf_` - let's put it in the base class and do this check in it's constructor.

src/daemon/emitter.cc
```diff
@@ -108,20 +110,30 @@ Emitter::Emitter(const proto::Configuration& configuration)
     }
   }
 
-  for (const auto& remote : config->emitter().remotes()) {
-    if (!remote.disabled()) {
+  if (runs_coordinators_task_) {
+    Vector<ResolveFn> resolvers;
+    resolvers.reserve(config->emitter().coordinators_size());
+    for (const auto& coordinator : config->emitter().coordinators()) {
+      if (coordinator.disabled())
```
abyss7: One-liners without braces are forbidden.

src/daemon/emitter.cc
```diff
@@ -108,20 +110,30 @@ Emitter::Emitter(const proto::Configuration& configuration)
...
         optional->Wait();
         return optional->GetValue();
       };
-      Worker worker = std::bind(&Emitter::DoRemoteExecute, this, _1, resolver);
-      workers_->AddWorker("Remote Execute Worker"_l, worker, remote.threads());
+      resolvers.push_back(resolver);
     }
+    Worker worker = std::bind(&Emitter::DoPoll, this, _1, resolvers);
+    workers_->AddWorker("Coordinator Poll Worker"_l, worker);
+  } else {
```
abyss7: By design we should at first consider the default remotes from config (if any), and only then try to use remotes from coordinators. It's better in case all coordinators are down, but we know some good remotes to fallback to.

src/daemon/emitter.cc
```diff
@@ -131,37 +143,65 @@ Emitter::~Emitter() {
   failed_tasks_->Close();
   local_tasks_->Close();
   workers_.reset();
+  // Create and destroy |compilation_workers_| pool on same thread.
```
abyss7: |remote_workers| ?

src/daemon/emitter.cc
```diff
@@ -203,6 +243,26 @@ void Emitter::SetExtraFiles(const cache::ExtraFiles& extra_files,
   }
 }
 
+void Emitter::SpawnCompilationWorkers(const proto::Configuration& config) {
```
abyss7: Let's rename to `SpawnRemoteWorkers`

src/daemon/emitter.cc
```diff
@@ -203,6 +243,26 @@ void Emitter::SetExtraFiles(const cache::ExtraFiles& extra_files,
...
+  using Worker = base::WorkerPool::SimpleWorker;
+  for (const auto& remote : config.emitter().remotes()) {
+    if (remote.disabled())
```
abyss7: One-liners without braces are forbidden.

abyss7: Please, review all other similar cases by your-self.

src/daemon/emitter.h
```diff
@@ -13,8 +13,11 @@ class Emitter : public CompilationDaemon {
   virtual ~Emitter();
 
   bool Initialize() override;
+  bool UpdateConfiguration(const proto::Configuration& configuration) override;
 
  private:
+  FRIEND_TEST(EmitterTest, GracefulConfigurationUpdate);
+  friend class CoordinatedTestEmitter;
```
abyss7: There should be no such class, doesn't it?

src/net/test_end_point.h
```diff
@@ -0,0 +1,21 @@
...
+
+#include <net/end_point.h>
+
+namespace dist_clang {
+namespace net {
+
+class TestEndPoint : public EndPoint {
+ public:
+  TestEndPoint(const String& host, ui16 port = 0u) : host_(host), port_(port) {}
+  ~TestEndPoint() override {}
```
abyss7: Is it required?

src/base/empty_lambda.h
```diff
@@ -11,7 +11,7 @@ class EmptyLambda {
 
   template <class... Args>
   ReturnType operator()(Args&&...) {
-    return ReturnType();
+    return value_;
```
matthewtff: Now it returns a value provided by user, not a default value for type.
I use it with a |true| value, but without this fix it returns |false|.

src/daemon/common_daemon_test.h
```diff
@@ -27,12 +29,15 @@ struct CommonDaemonTest : public ::testing::Test {
         test_service = service;
         service->CountConnectAttempts(&connect_count);
         service->CountListenAttempts(&listen_count);
-        service->CallOnConnect([this](net::EndPointPtr, String*) {
+        service_callback(service);
```
matthewtff: Added service_callback to run code once network_server was created, but before any other code had a chance to run. Used to call for Listen to imitate remote servers.
Otherwise races could occur as code run from daemon could reach to remote before test had a chance to call test_service methods.

src/daemon/clangd_main.cc
```diff
@@ -50,12 +51,26 @@ int main(int argc, char* argv[]) {
...
 
-  if (collector && !collector->Initialize()) {
-    LOG(FATAL) << "Collector failed to initialize.";
+  // Initialize coordinator, if any.
+  {
+    if (configuration.config().has_coordinator()) {
+      coordinator.reset(new daemon::Coordinator(configuration.config()));
+    }
+
+    if (coordinator && coordinator->Initialize()) {
```
matthewtff: It should be `!coordinator->Initialize()`.

matthewtff:
A new one with a rebase to actual master would be created

PR #91, opened 2016-12-16
Some refactoring

abyss7:
Move `conf_` to `BaseDaemon` since everyone uses it - and make it thread-safe.

Also implement testable `EndPoint` - for future tests with `Connect()`.

Original author is @matthewtff


matthewtff:
There's only 1 file changed in commit and has nothing about |conf_|

abyss7:
I have rewritten the history with proper commit.

PR #92, opened 2016-12-16
Move `conf_` to `BaseDaemon` since everyone uses it.

abyss7:
Make `conf_` thread-safe.
Implement testable `EndPoint` - for future tests with `Connect()`.

Original author is @matthewttf


abyss7:
Another attempt - in the first PR I forgot to commit all unstaged files.

PR #94, opened 2017-01-10
Added distribution support to LockedQueue

matthewtff:
Now Emitter finds out a hash of tasks for remote execution
and puts this info to LockedQueue. Also DoRemoteExecute now
has assigned distribution value that is used to pop appropriate
task from queue.

### DIFF
--- a/build/travis.sh
+++ b/build/travis.sh
@@ -14,7 +14,7 @@ elif [[ "$(uname)" == Darwin ]]; then
     arch="Mac"
 fi
 
-clang_revision=283753-1
+clang_revision=282487-1
 clang_root="$root_dir/out/clang-$clang_revision"
 if ! test -d "$clang_root"; then
     mkdir "$clang_root"
--- a/src/base/locked_queue.h
+++ b/src/base/locked_queue.h
@@ -5,7 +5,7 @@
 
 #include STL(condition_variable)
 #include STL(experimental/optional)
-#include STL(queue)
+#include STL(list)
 
 namespace dist_clang {
 namespace base {
@@ -17,6 +17,18 @@ template <class T>
 class LockedQueue {
  public:
   using Optional = std::experimental::optional<T>;
+  using Queue = List<T>;
+  enum InfoIndex {
+    ITERATOR = 0u,
+    HASH = 1u,
+  };
+  enum ReverseInfoIndex {
+    RANGE = 0u,
+    REVERSE_ITERATOR = 1u,
+  };
+  using TaskInfo = Tuple<typename Queue::iterator, size_t>;
+  // List of tasks from same 'range' in terms of distributed tasks.
+  using RangeList = List<TaskInfo>;
 
   enum {
     UNLIMITED = 0,
@@ -24,7 +36,18 @@ class LockedQueue {
 
   LockedQueue() = default;
   explicit LockedQueue(ui32 capacity) : capacity_(capacity) {}
-  ~LockedQueue() { DCHECK(closed_); }
+  ~LockedQueue() {
+    DCHECK(closed_);
+#if !defined(NDEBUG)
+    // Make sure we don't leak tasks.
+    DCHECK(reverse_index_.size() == queue_.size());
+    size_t index_size = 0u;
+    for (const RangeList& range_list : index_) {
+      index_size += range_list.size();
+    }
+    DCHECK(index_size == queue_.size());
+#endif  // !defined(NDEBUG)
+  }
 
   // Should be explicitly closed before destruction.
   void Close() THREAD_SAFE {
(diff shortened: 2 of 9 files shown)


src/base/worker_pool.h
```diff
@@ -25,6 +24,10 @@ class WorkerPool {
     return WaitUntilShutdown(ZERO_DURATION);
   }
 
+  size_t GetWorkersNumber() const {
```
matthewtff: Think it's unused now, should be removed.

abyss7:
Use common name for "shard" across the project, instead of "range", "distribution", etc.

abyss7:
Rebase onto new master.

abyss7:
Use preprocessor hash for sharding, instead of hash from input file basename.

abyss7:
Implement some tests:
1. Check that task goes to the right shard (not by random pick from a queue head).
2. Check that next same task goes to the same shard (not by random pick from a queue head).
3. Check that shards get updated after update from coordinator.

build/travis.sh
```diff
@@ -14,7 +14,7 @@ elif [[ "$(uname)" == Darwin ]]; then
     arch="Mac"
 fi
 
-clang_revision=283753-1
+clang_revision=282487-1
```
abyss7: Why to downgrade the version?

src/base/locked_queue.h
```diff
@@ -56,41 +63,130 @@ class LockedQueue {
...
+    }
+
+    size_t Size() const THREAD_UNSAFE {
+      size_t index_size = 0u;
+      for (const ShardsList& shards_list : index_) {
+        index_size += shards_list.size();
+      }
+      return index_size;
+    }
+    void Put(typename Queue::iterator item, const ui64 hash) {
```
abyss7: `hash` → `shard` ?

src/daemon/emitter.cc
```diff
@@ -394,18 +386,7 @@ void Emitter::DoRemoteExecute(const base::WorkerPool& pool,
...
-      std::get<CONNECTION>(*task)->ReportStatus(status);
-      continue;
-    }
-
-    auto outgoing = std::make_unique<proto::Remote>();
-    if (source.str.empty() && !GenerateSource(incoming, &source)) {
-      failed_tasks_->Push(std::move(*task));
-      continue;
-    }
+    CHECK(!source.str.empty());
```
abyss7: It's a very strong condition. What will happen, if we don't have local cache?

src/daemon/emitter.cc
```diff
@@ -653,13 +640,46 @@ bool Emitter::Reload(const proto::Configuration& conf) {
       optional->Wait();
       return optional->GetValue();
     };
-    Worker worker = std::bind(&Emitter::DoRemoteExecute, this, _1, resolver);
+    const ui32 shard =
+        remote.has_distribution() ? remote.distribution() : 0u;
```
abyss7: `distribution` → `shard` ?

src/daemon/emitter.h
```diff
@@ -24,11 +24,12 @@ class Emitter : public CompilationDaemon {
     MESSAGE = 1,
     SOURCE = 2,
     EXTRA_FILES = 3,
+    HASH = 4,
```
abyss7: `HASH` → `SHARD` ?

build/travis.sh
matthewtff: My bad, I'll revert this change

src/base/locked_queue.h
```diff
@@ -56,41 +63,130 @@ class LockedQueue {
...
+    enum InfoIndex {
+      ITERATOR = 0u,
+      HASH = 1u,
+    };
+    enum ReverseInfoIndex {
+      RANGE = 0u,
+      REVERSE_ITERATOR = 1u,
+    };
+
+    Index(LockedQueue<T>* const queue) : queue_(queue) {
```
matthewtff: That should be ok, yeah.

src/base/locked_queue.h
matthewtff: Actually, no. Here and in other places we hash is a hash value computed out of sources. Base on hash value and number of shards a shard could be computed. But anyway we need 2 distinct entities - shards and hashes.

src/daemon/emitter.cc
matthewtff: That would also mean a breakage in config files. Is it OK to check both for several releases and remove then?

src/daemon/emitter.h
matthewtff: Same as in other places about hash -> shard:
It's a calculated hash value, and shard should be computed out of this and current number of shards(that may change).

src/daemon/emitter.cc
matthewtff: It should be here in any case. To put to `all_tasks_` we need to now hash of sources of the task. All these fields are filled in `PopulateTask`. There should be no way to put a task to queue without knowing it's hash/sources.

PR #95, opened 2017-01-12
Refactor the configuration update mechanics

abyss7:
Also try to get rid of 'new' operator.
Minor fixes for the coordinator polling from emitter.
Some tests don't work right now.

### DIFF
--- a/src/daemon/absorber.cc
+++ b/src/daemon/absorber.cc
@@ -15,19 +15,17 @@ using namespace std::placeholders;
 namespace dist_clang {
 namespace daemon {
 
-Absorber::Absorber(const proto::Configuration& configuration)
-    : CompilationDaemon(configuration) {
+Absorber::Absorber(const Configuration& conf) : CompilationDaemon(conf) {
   using Worker = base::WorkerPool::SimpleWorker;
-  auto config = conf();
-  CHECK(config->has_absorber() && !config->absorber().local().disabled());
+  CHECK(conf.has_absorber() && !conf.absorber().local().disabled());
 
-  workers_.reset(new base::WorkerPool);
-  tasks_.reset(new Queue(config->pool_capacity()));
+  workers_ = std::make_unique<base::WorkerPool>();
+  tasks_ = std::make_unique<Queue>(conf.pool_capacity());
 
   {
     Worker worker = std::bind(&Absorber::DoExecute, this, _1);
     workers_->AddWorker("Execute Worker"_l, worker,
-                        config->absorber().local().threads());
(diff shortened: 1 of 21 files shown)


abyss7:
@matthewtff please take look - to be aware of upcoming changes.

src/daemon/base_daemon.h
```diff
@@ -10,22 +10,42 @@
 namespace dist_clang {
 namespace daemon {
 
+/** Daemon and configuration.
```
abyss7: The description will get fixed - it's outdated.

src/daemon/absorber.cc
```diff
@@ -37,16 +35,17 @@ Absorber::~Absorber() {
 }
 
 bool Absorber::Initialize() {
+  auto conf = this->conf();
```
matthewtff: What's the purpose of explicit `this` here and in other similar places?

src/daemon/compilation_daemon.h
```diff
@@ -51,6 +50,11 @@ class CompilationDaemon : public BaseDaemon {
                          const cache::ExtraFiles& extra_files,
                          const cache::FileCache::Entry& entry);
 
+  bool Check(const Configuration& conf) const override;
+  inline bool Reload(const Configuration& conf) override {
```
matthewtff: Why override to call base implementation?

src/daemon/emitter.cc
```diff
@@ -595,12 +534,16 @@ void Emitter::DoPoll(const base::WorkerPool& pool,
     }
   };
 
-  do {
+  auto conf = this->conf();
+  const auto poll_interval = conf->emitter().poll_interval();
+
+  while (!pool.WaitUntilShutdown(std::chrono::seconds(poll_interval))) {
```
matthewtff: When switching `do - while` to `while` emitter may wait too long for getting remotes from coordinator, isn't it?
If `poll_interval` equals 30 mins, for example.

abyss7: Reasonable. I'll change back.

src/daemon/absorber.cc
abyss7: Otherwise compiler gives an error about definition of `auto` variable. Also it's a little syntax trick to prevent implicit usage of raw `conf()`.

src/daemon/compilation_daemon.h
abyss7: For the purpose of uniformity of chained calls inside further overridden methods. Since I can't just invoke "the first defined same method in the nearest base class".

PR #96, opened 2017-02-02
Implement a sharding of tasks between remote hosts

abyss7:
Forked from @matthewtff branch

### DIFF
--- a/src/base/BUILD.gn
+++ b/src/base/BUILD.gn
@@ -46,6 +46,7 @@ shared_library("base") {
     "future.h",
     "locked_list.h",
     "locked_queue.h",
+    "locked_queue_index.h",
     "process.cc",
     "process.h",
     "process_forward.h",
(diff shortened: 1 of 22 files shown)


src/base/locked_queue.h
```diff
@@ -2,29 +2,40 @@
...
 
 template <class T>
 class QueueAggregator;
 
+// If the queue is shardless then by default it uses only zero shard.
 template <class T>
 class LockedQueue {
  public:
   using Optional = std::experimental::optional<T>;
+  using Queue = List<T>;
```
abyss7: No, since the `List` is visible anyway.

src/base/BUILD.gn
```diff
@@ -46,6 +46,7 @@ shared_library("base") {
     "future.h",
     "locked_list.h",
     "locked_queue.h",
+    "locked_queue_index.h",
```
abyss7: Done.

PR #102, opened 2017-05-11
Little perf fix for ConstString

matthewtff:
To access element by it's index introduced a private method 'at', which
requires a guarantee that |internals_| pointer wouldn't change
meanwhile.
This speeds up |ConstString::find| for about x10 as it makes *a lot* of
index lookups.

### DIFF
--- a/src/base/const_string.cc
+++ b/src/base/const_string.cc
@@ -196,7 +196,7 @@ bool ConstString::operator==(const ConstString& other) const {
 }
 
 size_t ConstString::find(const char* str) const {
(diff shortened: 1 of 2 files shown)


src/base/const_string.cc
```diff
@@ -213,8 +213,14 @@ size_t ConstString::find(const char* str) const {
 
   size_t sp = 0;
   i64 kp = 0;
+
+  // |internals_| may release in other thread, so create own copy of owning
```
abyss7: I think those comments are redundant. Maybe remove them for now?

PR #103, opened 2017-05-11
Support PCH & PTH files

matthewtff:
Use contents of PCH and PTH files to calculate hash of direct cache
entry.

### DIFF
--- a/src/cache/file_cache_test.cc
+++ b/src/cache/file_cache_test.cc
@@ -386,7 +386,7 @@ TEST(FileCacheTest, RestoreDirectEntry) {
   // Store the direct entry.
   const UnhandledSource orig_code("int main() {}"_l);
   const List<String> headers = {header1_path, header2_rel_path};
-  cache.Store(orig_code, {}, cl, version, headers, path,
+  cache.Store(orig_code, {}, cl, version, headers, {}, path,
               FileCache::Hash(code, {}, cl, version));
 
   // Restore the entry.
@@ -435,7 +435,7 @@ TEST(FileCacheTest, RestoreDirectEntryWithExtraFile) {
   const List<String> headers = {header1_path, header2_rel_path};
   cache.Store(
       orig_code, ExtraFiles{{SANITIZE_BLACKLIST, extra_file}}, cl, version,
-      headers, path,
+      headers, {}, path,
       FileCache::Hash(code, ExtraFiles{{SANITIZE_BLACKLIST, extra_file}}, cl,
                       version));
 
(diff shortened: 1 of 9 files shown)


src/client/clang_command.cc
```diff
@@ -100,6 +100,11 @@ bool ClangCommand::FillFlags(base::proto::Flags* flags,
              arg->getOption().matches(OPT_D) ||
              arg->getOption().matches(OPT_I)) {
       arg->render(arg_list_, non_cached_list);
+    } else if (arg->getOption().matches(OPT_include_pch) ||
+               arg->getOption().matches(OPT_include_pth)) {
+      auto* included_file = flags->add_included_files();
```
abyss7: Maybe just `flags->add_included_files(arg->getValue())` ?

src/client/clang_command.cc
```diff
@@ -100,6 +100,11 @@ bool ClangCommand::FillFlags(base::proto::Flags* flags,
...
+      *included_file = arg->getValue();
+      arg->render(arg_list_, non_cached_list);
```
abyss7: Why to add to direct cache an explicit argument? - It's already taken into account in flags as `included_files`. You should render them later, like `sanitize_blacklist` or `deps_file`.

src/client/command.cc
```diff
@@ -97,9 +97,8 @@ bool Command::GenerateFromArgs(int argc, const char* const raw_argv[],
     // TODO: move this code inside |DriverCommand| constructor.
     if ((command.getSource().getKind() != Action::AssembleJobClass &&
          command.getSource().getKind() != Action::BackendJobClass &&
-         command.getSource().getKind() != Action::CompileJobClass &&
-         command.getSource().getKind() != Action::PrecompileJobClass &&
```
abyss7: Why do we skip the precompile jobs?

src/client/clang_command.cc
```diff
@@ -100,6 +100,11 @@ bool ClangCommand::FillFlags(base::proto::Flags* flags,
              arg->getOption().matches(OPT_D) ||
              arg->getOption().matches(OPT_I)) {
       arg->render(arg_list_, non_cached_list);
+    } else if (arg->getOption().matches(OPT_include_pch) ||
+               arg->getOption().matches(OPT_include_pth)) {
+      auto* included_file = flags->add_included_files();
+      *included_file = arg->getValue();
```
abyss7: What are the rules for path calculation for a value of this option? Is it relative to current folder or input file? Or does it use include paths?

matthewtff: It is calculated relative to current work directory.

src/client/clang_command.cc
matthewtff: This flag could be useful for preprocessing the source code, for example.
Why bother with manual rendering?

src/client/command.cc
matthewtff: As those work wrong on remotes. It looks like a PCH file generated on remote host doesn't work locally. There're may be several different reasons for this:
1. OS/Arch differencies(by order?) as pch/pth files are somethig like a memory dump.
2. These files contain absolute paths to source header.

src/client/clang_command.cc
abyss7: Even without manual rendering still I don't see the point to render it inside `non_cached_list` - it's already counted inside `manifest.headers` for direct cache, and `non_cached_list` is not applied for simple cache.

src/cache/file_cache_test.cc
```diff
@@ -496,6 +496,53 @@ TEST(FileCacheTest, DirectEntry_ChangedHeaderContents) {
...
 }
 
+TEST(FileCacheTest, DirectEntry_ChangedPreprocessedHeaderContents) {
+  const base::TemporaryDir tmp_dir;
+  const String path = tmp_dir;
+  const String object_path = path + "/test.o";
+  const String deps_path = path + "/test.d";
+  const String header1_path = path + "/test1.h";
+  const String header2_path = path + "/test2.h";
+  const String preprocessed_header_path = path + "/precompile.h";
```
abyss7: Should use real-life extension, like .pch or .pth

matthewtff:
Updated EmitterTest and used real life extensions.
https://github.com/abyss7/dist-clang/blob/master/src/cache/file_cache_test.cc#L31 makes no sense for preprocessed headers as they don't play in hash computation.

PR #105, opened 2017-05-18
Add several metric reports. Also enlarge a little tests coverage

matthewtff:


src/daemon/common_daemon_test.h
```diff
@@ -21,6 +22,7 @@ struct CommonDaemonTest : public ::testing::Test {
   using RunCallback = Fn<void(base::TestProcess*)>;
 
   void SetUp() override {
+    DumpStatisticValues();
```
abyss7: What is it for?

src/daemon/compilation_daemon.cc
```diff
@@ -243,6 +246,8 @@ bool CompilationDaemon::SearchSimpleCache(
   if (!cache_) {
     return false;
   }
+  perf::Counter<perf::StatReporter, true> counter(
```
abyss7: `perf::Counter<perf::StatReporter>` should be fine everywhere, since the template argument is `true` by default.

src/daemon/emitter.cc
```diff
@@ -332,6 +334,8 @@ void Emitter::DoLocalExecute(const base::WorkerPool& pool) {
 
     ui32 uid =
         incoming->has_user_id() ? incoming->user_id() : base::Process::SAME_UID;
+    perf::Counter<perf::StatReporter, true> counter(
```
abyss7: Do you want to double-count the cache update as a local compilation time too?

abyss7: Do you want to count the failed compilations too?

src/daemon/emitter.cc
```diff
@@ -478,10 +494,13 @@ void Emitter::DoRemoteExecute(const base::WorkerPool& pool, ResolveFn resolver,
       if (status.code() != net::proto::Status::OK) {
         LOG(WARNING) << "Remote compilation failed with error(s):" << std::endl
                      << status.description();
+        STAT(REMOTE_COMPILATION_FAILED);
         failed_tasks_->Push(std::move(*task));
         counter.ReportOnDestroy(true);
         continue;
       }
+    } else {
+      STAT(REMOTE_COMPILATION_REJECTED);
```
abyss7: There is `OVERLOAD` in status codes, but it's unused for now. I suggest to implement the overload reporting in a proper way on absorber.

src/daemon/absorber.cc
```diff
@@ -45,7 +45,8 @@ bool Absorber::Initialize() {
     return false;
   }
 
-  if (conf->has_cache() && conf->cache().has_direct()) {
+  if (conf->has_cache() && conf->cache().has_direct() &&
+      conf->cache().direct()) {
```
matthewtff: Now absorber puts to log this message if it contains field `cache.direct`, even if it equals `false`. In this case message makes no sence. With this change message appears only if `cache.direct` equals `true`(which can't be used on absorber).

src/daemon/common_daemon_test.h
matthewtff: GTEST runs tests in same process if you don't ask it to spawn separate process for every new test. It means that several tests share same global objects like PerfService. To make results of one test not mess with results of other one it would be nice to clear results before every test run.

src/daemon/emitter.cc
matthewtff: No. I'll fix this issue, thanks.

matthewtff: Yep.

src/daemon/emitter.cc
matthewtff: Ah, I see. OK, I'll do it.

matthewtff:
Fixed local & remote(also included cache updates) compilation timings.
Also added an `OVERLOADED` status for absorber and take that change into account on emitter. Also updated tests.

src/daemon/emitter_test.cc
```diff
@@ -1837,6 +1859,431 @@ TEST_F(EmitterTest, StoreSimpleCacheForRemoteResult) {
...
+        const auto& command = message.GetExtension(proto::Remote::extension);
+        EXPECT_EQ(source, command.source());
+
+        send_condition.notify_all();
+      });
+      connection->CallOnRead([&](net::Connection::Message* message) {
+        message->MutableExtension(net::proto::Status::extension)
+            ->set_code(net::proto::Status::EXECUTION);
+      });
+    } else if (connect_count == 3) {
```
abyss7: Why do you need the 3rd connection in this test?

src/daemon/emitter_test.cc
```diff
@@ -1837,6 +1859,431 @@ TEST_F(EmitterTest, StoreSimpleCacheForRemoteResult) {
...
+  };
+
+  run_callback = [&](base::TestProcess* process) {
+    if (run_count == 1) {
+      EXPECT_EQ((Immutable::Rope{"-E"_l, "-x"_l, language, "-o"_l, "-"_l,
+                                 input_path1}),
+                process->args_);
+      process->stdout_ = source;
+    } else if (run_count == 2) {
+      // Used to check preprocess timing statistics.
```
abyss7: Copy-paste? `preprocess`?

src/daemon/emitter_test.cc
```diff
@@ -1837,6 +1859,431 @@ TEST_F(EmitterTest, StoreSimpleCacheForRemoteResult) {
...
+        const auto& command = message.GetExtension(proto::Remote::extension);
+        EXPECT_EQ(source, command.source());
+
+        send_condition.notify_all();
+      });
+      connection->CallOnRead([&](net::Connection::Message* message) {
+        message->MutableExtension(net::proto::Status::extension)
+            ->set_code(net::proto::Status::OVERLOAD);
+      });
+    } else if (connect_count == 3) {
```
abyss7: Same question: what is the 3rd connection for?

PR #107, opened 2017-05-30
Do not rewrite clangd.conf on deb package install/update

matthewtff:


matthewtff:
Based on https://wiki.debian.org/ConfigPackages

install/debian_postinst
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+
+set -e
+PKG=dist-clang
+if [ "$1" = configure ] ; then
+        for f in clangd.conf
```
abyss7: Why to use `for` loop for one element?

matthewtff: To stick as close to code from debian wiki as possible. As I'm not a shell hacker :(

I could change to single variable if you wish, but this version seems more extensible to me.

PR #110, opened 2017-06-09
Implemented strict sharding

matthewtff:


abyss7:
Will you do anything with tests?

matthewtff:
OMG, I missed tests failures. I'll fix them soon, thanks.

matthewtff:
Looks like I was buzz when coded those tests =/
Updated with the right version.

src/daemon/emitter.cc
```diff
@@ -453,9 +472,18 @@ void Emitter::DoRemoteExecute(const base::WorkerPool& pool, ResolveFn resolver,
         counter.ReportOnDestroy(false);
         LOG(WARNING) << "Failed to connect to " << end_point->Print() << ": "
                      << error;
-        // Put into |failed_tasks_| to prevent hanging around in case all
-        // remotes are unreachable at once.
-        failed_tasks_->Push(std::move(*task));
+        bool& shard_switched = std::get<CHANGED_SHARD>(*task);
+        if (conf->emitter().has_total_shards() &&
+            conf->emitter().shard_queue_limit() && !shard_switched) {
+          // Let other shard complete task on connection failure. Do it once.
```
abyss7: Describe this logic better in comments: at least about `shard_queue_limit()` indicates enabled strict sharding that prevents this task from completion in case the remote server has gone.

src/daemon/emitter.cc
```diff
@@ -579,6 +608,27 @@ void Emitter::DoRemoteExecute(const base::WorkerPool& pool, ResolveFn resolver,
     failed_tasks_->Push(std::move(*task));
     counter.ReportOnDestroy(true);
   }
+
+  auto new_conf = this->conf();
```
abyss7: I have to verify that we can guarantee the configuration is already reloaded at this point

src/daemon/emitter.cc
```diff
@@ -579,6 +608,27 @@ void Emitter::DoRemoteExecute(const base::WorkerPool& pool, ResolveFn resolver,
...
+
+  // In case if pool was shut down to apply new configuration check if
+  // new number of shards is smaller than our shard. If so, take out tasks
+  // from our shard and distribute them across other shards.
+  if (!all_tasks_->IsClosed() &&
+      new_conf->emitter().shard_queue_limit() > 0 &&
```
abyss7: extra `> 0`

src/daemon/emitter.cc
```diff
@@ -579,6 +608,27 @@ void Emitter::DoRemoteExecute(const base::WorkerPool& pool, ResolveFn resolver,
     failed_tasks_->Push(std::move(*task));
     counter.ReportOnDestroy(true);
   }
+
+  auto new_conf = this->conf();
+
+  // In case if pool was shut down to apply new configuration check if
+  // new number of shards is smaller than our shard. If so, take out tasks
+  // from our shard and distribute them across other shards.
+  if (!all_tasks_->IsClosed() &&
```
abyss7: Why to check if queue is closed?

src/daemon/emitter.cc
```diff
@@ -579,6 +608,27 @@ void Emitter::DoRemoteExecute(const base::WorkerPool& pool, ResolveFn resolver,
...
+      new_conf->emitter().shard_queue_limit() > 0 &&
+      new_conf->emitter().total_shards() <= shard) {
+    bool shard_is_empty = false;
+    do {
+      Optional&& task =
+          all_tasks_->Pop(pool, conf->emitter().shard_queue_limit(), shard);
```
abyss7: How do we guarantee that we won't start popping tasks from other shards?

abyss7: Why using `conf`? - Previous configuration may lack the strict sharding property.

src/daemon/emitter.cc
```diff
@@ -22,6 +24,17 @@ using Counter = perf::Counter<perf::StatReporter, ReportByDefault>;
 
 namespace {
 
+// Select a new shard, different from current.
+inline ui32 FindNewShard(std::random_device& random_device,
```
abyss7: Why not to incapsulate random device as static variable?

src/daemon/configuration.proto
```diff
@@ -42,20 +42,25 @@ message Configuration {
...
+    optional uint32 total_shards    = 7;
     // Number of total shards can't be less than 2.
 
-    optional uint32 pop_timeout   = 8 [ default = 1 ];
+    optional uint32 pop_timeout     = 8 [ default = 1 ];
     // in seconds - can't be zero.
+
+    optional uint32 shard_queue_limit = 9 [ default = 0 ];
+    // Implies strict sharding if specified limit is larger than zero.
+    // All tasks of shard that exceed limit can be destributed between other
```
abyss7: distributed -> redistributed?

src/daemon/emitter.cc
```diff
@@ -391,6 +406,7 @@ void Emitter::DoLocalExecute(const base::WorkerPool& pool) {
 void Emitter::DoRemoteExecute(const base::WorkerPool& pool, ResolveFn resolver,
                               const ui32 shard) {
   auto conf = this->conf();
+  std::random_device random_device;
```
matthewtff: AFAICT `/dev/urandom` is the preferred option in this case, instead of `/dev/random` as latter may block, while urandom never blocks. Also `/dev/urandom` is a default source used in libc++.

src/daemon/emitter.cc
```diff
@@ -430,15 +447,17 @@ void Emitter::DoRemoteExecute(const base::WorkerPool& pool, ResolveFn resolver,
 
     // Check that we have a compiler of a requested version.
     net::proto::Status status;
-    if (!SetupCompiler(incoming->mutable_flags(), &status)) {
+    const bool has_compiler_path =
```
matthewtff: We may come here two times(for initial shard and once again if shard was down). So to prevent "setting up compiler" twice, we check if it was already setted by checking flags of incoming task.

src/daemon/emitter.cc
matthewtff: It's not safe to use same `std::random_device` from different threads.

src/daemon/emitter.cc
matthewtff: Pool gets shutted down only after a new pool(with new configuration) has started. So yeah, it should've beed reloaded by this time.

src/daemon/emitter.cc
matthewtff: Then the default value(`0`) will be used. This means that all tasks without shards will be popped and distributed across new shards. Looks like a valid behavoir to me.

src/daemon/emitter.cc
matthewtff: I've added this to check if new configuration supports/expects sharding.
There might be cases with updates of configuration without shards to another configuration without shards.
In this case there's nothing to be done.

src/daemon/emitter.cc
matthewtff: To skip useless work. Not critical, but may speedup(not slowdown) daemon shutdown, for example.

src/daemon/emitter.cc
abyss7: Use [thread_local](http://en.cppreference.com/w/cpp/language/storage_duration)

src/daemon/emitter.cc
abyss7: Are there no options for PRNG?

src/daemon/emitter.cc
abyss7: Then why don't you move this logic inside `SetupCompiler()` and handle it properly and transparently? - Otherwise you risk to end up with a different logic in different places.

matthewtff: Ah, good point. I'll do that, thanks.

src/daemon/emitter.cc
matthewtff: What kind of options do you want?) There're different kind of distributions, but random device doesn't depend on any such thing.
The only option I see is initial seed.

src/base/locked_queue_index.h
```diff
@@ -19,46 +26,108 @@ class LockedQueue<T>::Index {
...
+    return index_[shard].tasks.empty();
+  }
+
+  void WaitForShard(const ui32 shard, UniqueLock& lock,
+                    const Seconds& timeout) THREAD_UNSAFE {
+    CHECK(shard < index_.size());
+    index_[shard].pop_condition_.wait_for(lock, timeout);
+  }
+
+  void NotifyShard(const ui32 shard) THREAD_UNSAFE {
```
abyss7: You should review all your `THREAD_UNSAFE` attributes, since at least this method is used in thread-safe `Push()` without locks.

src/base/locked_queue.h
```diff
@@ -62,6 +65,7 @@ class LockedQueue {
     }
     ++size_;
     pop_condition_.notify_one();
+    index_.NotifyShard(shard);
```
abyss7: Why to notify twice? How it should work?

src/base/locked_queue.h
```diff
@@ -83,28 +87,57 @@ class LockedQueue {
...
+  }
 
-    auto it = index_.Get(shard, queue_.begin());
-    Optional&& obj = std::move(*it);
-    queue_.erase(it);
+  Optional PopStrict(const WorkerPool& pool,
+                     UniqueLock& lock,
+                     ui32 shard = DEFAULT_SHARD) THREAD_SAFE {
+    do {
+      index_.WaitForShard(shard, lock, timeout_);
```
abyss7: That's not good in case you've got overloaded shard. It may become not overloaded just before this call - and you will waste time for irrelevant shard.

src/base/locked_queue.h
```diff
@@ -83,28 +87,57 @@ class LockedQueue {
...
+    } while (!closed_ && index_.ShardIsEmpty(shard) && !pool.IsShuttingDown());
+    if (closed_ || index_.ShardIsEmpty(shard)) {
```
abyss7: Why don't we pop remaining tasks after close - like in `PopWithHint()`?

src/base/locked_queue_index.h
```diff
@@ -19,46 +26,108 @@ class LockedQueue<T>::Index {
...
-      index_.resize(shard + 1);
+  void Put(QueueIterator it, const ui32 shard) {
+    EnsureShardExists(shard);
+    reverse_index_[&*it] = std::make_pair(
+        index_[shard].tasks.insert(index_[shard].tasks.end(), it),
+        shard);
+  }
+
+  bool ShardIsEmpty(const ui32 shard) THREAD_UNSAFE {
+    CHECK(shard < index_.size());
```
abyss7: Why to use CHECKs? It's expensive to call `size()` in production code, IMHO.

src/base/locked_queue_index.h
```diff
@@ -19,46 +26,108 @@ class LockedQueue<T>::Index {
...
+    index_[shard].pop_condition_.wait_for(lock, timeout);
+  }
+
+  void NotifyShard(const ui32 shard) THREAD_UNSAFE {
+    CHECK(shard < index_.size());
+    index_[shard].pop_condition_.notify_one();
+  }
+
+  // Returns either |shard| or shard with queue larger than |shard_queue_limit|.
+  ui32 MaybeOverloadedShard(const ui32 shard_queue_limit,
```
abyss7: `DCHECK(shard_queue_limit != NOT_STRICT_SHARDING)` ?

src/base/locked_queue_index.h
```diff
@@ -19,46 +26,108 @@ class LockedQueue<T>::Index {
...
+                            const ui32 shard) THREAD_UNSAFE {
+    EnsureShardExists(shard);
+    if (index_[shard].tasks.size() > shard_queue_limit) {
```
abyss7: Why not to prefer initial shard if it has at least one task?

src/base/locked_queue_index.h
```diff
@@ -19,46 +26,108 @@ class LockedQueue<T>::Index {
...
+      return shard;
     }
-    reverse_index_[&*it] =
-        std::make_pair(index_[shard].insert(index_[shard].end(), it), shard);
+
+    ui32 overloaded_shard = 0u;
+    ui32 overloaded_shard_queue_size = 0u;
+    for (ui32 shard = 0; shard < index_.size(); ++shard) {
```
abyss7: Why not just take the first overloaded shard?

abyss7:
Tests have failed

matthewtff:
Don't pay attention to that push, I didn't mean to update PR :(

I'll ping you once done.

src/daemon/emitter.cc
matthewtff: I believe we could pass `std::numeric_limits<ui32>::max()` for `shard_queue_limit` here.

src/base/locked_queue.h
```diff
@@ -61,56 +65,111 @@ class LockedQueue {
...
+        index_.MaybeOverloadedShard(shard_queue_limit, shard);
+    if (shard != maybe_overloaded_shard) {
+      return RemoveTaskFromQueue(index_.GetStrict(maybe_overloaded_shard));
+    }
 
-    auto it = index_.Get(shard, queue_.begin());
-    Optional&& obj = std::move(*it);
-    queue_.erase(it);
+
+    // See comment about while look in |PopWithHint|.
```
abyss7: while loop ?

src/base/locked_queue.h
```diff
@@ -61,56 +65,111 @@ class LockedQueue {
...
       return Optional();
     }
+    return RemoveTaskFromQueue(index_.GetWithHint(shard, queue_.begin()));
+  }
+
+  Optional PopStrict(const WorkerPool& pool,
+                     UniqueLock& lock,
+                     const ui32 shard_queue_limit,
+                     const ui32 shard) THREAD_SAFE {
+    // Check if there's a shard with number of tasks that exceed queue limit
```
abyss7: Add `DCHECK`'s for `shard_queue_limit` and `shard`.

src/base/locked_queue.h
```diff
@@ -61,56 +65,111 @@ class LockedQueue {
...
+    // AND current shard has no tasks to do. If so, steal a task from overloaded
+    // shard.
+    const ui32 maybe_overloaded_shard =
+        index_.MaybeOverloadedShard(shard_queue_limit, shard);
+    if (shard != maybe_overloaded_shard) {
```
abyss7: Why not to try more overloaded shards, til they exist?

src/base/locked_queue.h
```diff
@@ -61,56 +65,111 @@ class LockedQueue {
...
+    // One can't wait for condition using predicate and timed wait here as once
+    // the waiting timed out, the |pool.IsShuttingDown()| should be checked and
+    // wait again if pool doesn't shutting down.
+    while (!closed_ && queue_.empty() && !pool.IsShuttingDown()) {
+      index_.WaitShardFor(shard, lock, timeout_);
+    }
     if ((closed_ && queue_.empty()) || pool.IsShuttingDown()) {
       return Optional();
     }
+    return RemoveTaskFromQueue(index_.GetWithHint(shard, queue_.begin()));
```
abyss7: You should mark the method `PopWithHint` as `THREAD_UNSAFE` otherwise it won't work in multiple threads.

src/base/locked_queue.h
```diff
@@ -61,56 +65,111 @@ class LockedQueue {
...
+                     UniqueLock& lock,
+                     const ui32 shard_queue_limit,
+                     const ui32 shard) THREAD_SAFE {
+    // Check if there's a shard with number of tasks that exceed queue limit
+    // AND current shard has no tasks to do. If so, steal a task from overloaded
+    // shard.
+    const ui32 maybe_overloaded_shard =
+        index_.MaybeOverloadedShard(shard_queue_limit, shard);
+    if (shard != maybe_overloaded_shard) {
+      return RemoveTaskFromQueue(index_.GetStrict(maybe_overloaded_shard));
```
abyss7: You should mark the method `PopStrict` as `THREAD_UNSAFE` otherwise it won't work in multiple threads.

src/base/locked_queue_index.h
```diff
@@ -19,46 +29,115 @@ class LockedQueue<T>::Index {
...
+  }
+
+  template <typename Pred>
+  void WaitShard(const ui32 shard, UniqueLock& lock,
+                 const Pred& pred) THREAD_UNSAFE {
+    DCHECK(shard == LockedQueue<T>::DEFAULT_SHARD);
+    index_[shard].pop_condition_.wait(lock, pred);
+  }
+
+  void NotifyShard(const ui32 shard) THREAD_UNSAFE {
```
abyss7: Why is it `THREAD_UNSAFE`?

src/base/locked_queue_index.h
```diff
@@ -19,46 +29,115 @@ class LockedQueue<T>::Index {
...
+                    const Seconds& timeout) THREAD_UNSAFE {
+    EnsureShardExists(shard);
+    DCHECK(shard < index_.size());
+    index_[shard].pop_condition_.wait_for(lock, timeout);
+  }
+
+  template <typename Pred>
+  void WaitShard(const ui32 shard, UniqueLock& lock,
+                 const Pred& pred) THREAD_UNSAFE {
+    DCHECK(shard == LockedQueue<T>::DEFAULT_SHARD);
```
abyss7: Why to accept `shard` if it supposed to have always the same value?

src/base/locked_queue_test.cc
```diff
@@ -110,9 +113,85 @@ TEST(LockedQueueTest, UniquePtrFriendliness) {
...
+  const int total_number_of_tasks = number_of_tasks_to_process + tasks_to_leave;
+
+  const size_t initial_number_of_shards = 4;
+  UniquePtr<WorkerPool> workers(new WorkerPool);
+  for (size_t shard = 0; shard < initial_number_of_shards; ++shard) {
+    Worker worker = [shard, &queue](const WorkerPool& pool) {
+      const int tasks_per_shard =
+          number_of_tasks_to_process / initial_number_of_shards;
+      for (int task = 0; task < tasks_per_shard; ++task) {
+        EXPECT_TRUE(!!queue.Pop(pool, 0, shard));
```
abyss7: `0` → `NOT_STRICT_SHARDING`

src/base/locked_queue_test.cc
```diff
@@ -124,20 +203,20 @@ TEST(LockedQueueIndexTest, BasicUsage) {
...
   // Index shouldn't shrink on elements removal.
   EXPECT_EQ(shard + 1, index.index_.size());
-  EXPECT_EQ(0u, index.index_[shard].size());
+  EXPECT_EQ(0u, index.index_[shard].tasks.size());
   EXPECT_EQ(0u, index.reverse_index_.size());
   EXPECT_EQ(0u, index.reverse_index_.count(&*inserted_item));
 }
 
-TEST(LockedQueueIndexTest, GetFromHead) {
+TEST(LockedQueueIndexTest, GetWithHint) {
```
abyss7: `GetWithHintFromHead` ?

src/base/locked_queue_test.cc
```diff
@@ -147,20 +226,39 @@ TEST(LockedQueueIndexTest, GetFromHead) {
...
+
+  ASSERT_ANY_THROW(index.GetStrict(shard));
+
+  index.Put(inserted_item, shard);
+  EXPECT_EQ(shard + 1, index.index_.size());
+  EXPECT_EQ(1u, index.index_[shard].tasks.size());
+  EXPECT_EQ(1u, index.reverse_index_.size());
+  EXPECT_EQ(1u, index.reverse_index_.count(&*inserted_item));
+
+  EXPECT_EQ(inserted_item, index.GetStrict(shard));
```
abyss7: Why not to check that `GetStrict()` doesn't return from another shard?

src/base/locked_queue_test.cc
```diff
@@ -170,15 +268,71 @@ TEST(LockedQueueIndexTest, ShardIndexGrows) {
...
+  const ui32 max_queue_size = 3u;
+
+  EXPECT_EQ(shard2, index.MaybeOverloadedShard(max_queue_size, shard2));
+  EXPECT_EQ(shard2 + 1, index.index_.size());
+  EXPECT_EQ(1u, index.index_[shard1].tasks.size());
+  EXPECT_EQ(1u, index.reverse_index_.count(&*inserted_item));
+  EXPECT_EQ(0u, index.index_[shard2].tasks.size());
+}
+
+TEST(LockedQueueIndexTest, MaybeOverloadedReturnsOverloadedShard) {
```
abyss7: Don't forget to check that `MaybeOverloaded` considers initial shard non-emptiness

src/daemon/emitter.cc
```diff
@@ -618,18 +619,24 @@ void Emitter::DoRemoteExecute(const base::WorkerPool& pool, ResolveFn resolver,
...
-          all_tasks_->Pop(pool, conf->emitter().shard_queue_limit(), shard);
+          all_tasks_->Pop(pool, std::numeric_limits<ui32>::max(), shard);
       if (task) {
-        all_tasks_->Push(std::move(*task), FindNewShard(
-            new_conf->emitter().total_shards(), shard));
+        // Once we popped a valid task - put it to appropriate shard according
+        // to new number of shards.
+        auto& handled_hash = std::get<HANDLED_HASH>(*task);
+        const ui32 new_shard =
+            (*reinterpret_cast<const ui32*>(handled_hash.str.Hash(4).c_str()))
```
abyss7: That not a good place to duplicate the sharding algorithm

src/daemon/emitter_test.cc
```diff
@@ -284,12 +284,15 @@ TEST_F(EmitterTest, ConfigurationUpdateFromCoordinator) {
                 end_point->Print());
 
       connection->CallOnSend([this](const net::Connection::Message&) {
+
         send_condition.notify_all();
         // Run the task #1 on old remote before sending anything to coordinator.
 
         UniqueLock lock(send_mutex);
+        // Wait until emitter sends task to old remote:
+        //  * 1st seconds - wait for queue to pop in old remote
```
abyss7: 1st second ?

src/base/locked_queue.h
```diff
@@ -41,11 +44,14 @@ class LockedQueue {
...
     UniqueLock lock(pop_mutex_);
     closed_ = true;
-    pop_condition_.notify_all();
+    index_.NotifyAllShards();
+    aggregator_condition_.notify_all();
   }
 
   inline ui32 Size() const THREAD_SAFE { return size_; }
 
+  bool IsClosed() const THREAD_SAFE { return closed_; }
```
abyss7: Add `inline`

src/base/locked_queue.h
```diff
@@ -61,56 +67,112 @@ class LockedQueue {
...
+      return PopWithHint(pool, lock, shard);
+    } else {
+      return PopStrict(pool, lock, shard_queue_limit, shard);
+    }
+  }
+
+ private:
+  Optional PopWithHint(const WorkerPool& pool, UniqueLock& lock,
+                       const ui32 shard) THREAD_UNSAFE {
+    // One can't wait for condition using predicate and timed wait here as once
```
abyss7: "at once" ?

src/base/locked_queue_index.h
```diff
@@ -19,46 +29,112 @@ class LockedQueue<T>::Index {
...
+  }
+
+  template <typename Pred>
+  void WaitDefaultShard(UniqueLock& lock, const Pred& pred) THREAD_UNSAFE {
+    index_[LockedQueue<T>::DEFAULT_SHARD].pop_condition_.wait(lock, pred);
+  }
+
+  void NotifyShard(const ui32 shard) THREAD_SAFE {
+    DCHECK(shard < index_.size());
+    index_[shard].pop_condition_.notify_one();
```
abyss7: Since this method pretends to be thread-safe, how can we guarantee that the `size()` is changed atomically with `index_[shard]`?

src/base/locked_queue.h
```diff
@@ -61,56 +67,112 @@ class LockedQueue {
...
+        index_.MaybeOverloadedShard(shard_queue_limit, shard);
+    if (shard != maybe_overloaded_shard) {
+      return RemoveTaskFromQueue(index_.GetStrict(maybe_overloaded_shard));
+    }
 
-    auto it = index_.Get(shard, queue_.begin());
-    Optional&& obj = std::move(*it);
-    queue_.erase(it);
+    auto shard_is_closed = [&] {
+      return closed_ || shard_queue_limit == GET_SHARD_DOWN;
```
abyss7: Please, add comment with scenario for `shard_queue_limit` - I spent 15 minutes to understand it.

src/base/locked_queue_test.cc
```diff
@@ -170,30 +171,82 @@ TEST(LockedQueueTest, ShardsNumberGrow) {
...
+  UniquePtr<WorkerPool> workers(new WorkerPool);
+
+  for (ui32 task = 0u; task < number_of_tasks; ++task) {
+    EXPECT_TRUE(queue.Push(task, shard_with_tasks));
+  }
+
+  Worker worker = [&](const WorkerPool& pool) {
+    for (size_t task = 0u; task < available_tasks; ++task) {
+      EXPECT_TRUE(!!queue.Pop(pool, shard_queue_limit, shard_without_tasks));
+    }
```
abyss7: May be try to pop another task and fail on pool shutdown?

src/daemon/emitter.cc
```diff
@@ -790,10 +761,32 @@ bool Emitter::Check(const Configuration& conf) const {
...
   using Worker = base::WorkerPool::SimpleWorker;
 
-  if (conf.emitter().total_shards()) {
-    current_shard_number_ = conf.emitter().total_shards();
-  } else {
-    current_shard_number_ = 1u;
+  auto old_conf = this->conf();
+
+  // In case if new configurations honors strict sharding and has lower number
+  // of total shards, make sure tasks from abadonned tasks get redistributed
```
abyss7: "abandoned" ?

src/base/locked_queue_index.h
matthewtff: `index_` never shrinks. So if we verified that `shard` is lower than `size()`, then `index_[shard]` will always exist.

src/base/locked_queue.h
```diff
@@ -61,56 +67,120 @@ class LockedQueue {
...
+    // See a comment on |aggregator_condition_| to see why two notifications
+    // should be done here.
+    aggregator_condition_.notify_one();
 
     return true;
   }
 
   // Returns disengaged object only when this queue is closed and empty.
-  Optional Pop(ui32 shard = DEFAULT_SHARD) THREAD_SAFE {
+  Optional Pop() THREAD_SAFE {
```
matthewtff: Leave a comment about deprecating `ThreadPool`.

src/base/locked_queue.h
```diff
@@ -61,56 +67,120 @@ class LockedQueue {
...
+    return RemoveTaskFromQueue(
+        index_.GetWithHint(DEFAULT_SHARD, queue_.begin()));
   }
 
   // Returns disengaged object only when this queue is closed and empty, or pool
   // is shutting down.
-  Optional Pop(const WorkerPool& pool, ui32 shard = DEFAULT_SHARD) THREAD_SAFE {
+  Optional Pop(const WorkerPool& pool,
+               const ui32 shard_queue_limit = NOT_STRICT_SHARDING,
+               const ui32 shard = DEFAULT_SHARD) THREAD_SAFE {
```
abyss7: Add boolean flag about non-blocking behavior

src/base/locked_queue_index.h
```diff
@@ -19,46 +29,112 @@ class LockedQueue<T>::Index {
...
+    DCHECK(shard < index_.size());
+    index_[shard].pop_condition_.wait_for(lock, timeout);
+  }
+
+  template <typename Pred>
+  void WaitDefaultShard(UniqueLock& lock, const Pred& pred) THREAD_UNSAFE {
+    index_[LockedQueue<T>::DEFAULT_SHARD].pop_condition_.wait(lock, pred);
+  }
+
+  void NotifyShard(const ui32 shard) THREAD_SAFE {
```
matthewtff: Change to `THREAD_UNSAFE`

src/base/locked_queue_index.h
```diff
@@ -19,46 +29,112 @@ class LockedQueue<T>::Index {
...
-        std::make_pair(index_[shard].insert(index_[shard].end(), it), shard);
   }
 
-  QueueIterator Get(ui32 shard, QueueIterator begin) THREAD_UNSAFE {
+  // Returns either |shard| or shard with queue larger than |shard_queue_limit|.
+  ui32 MaybeOverloadedShard(const ui32 shard_queue_limit,
+                            const ui32 shard) THREAD_UNSAFE {
+    DCHECK(shard_queue_limit != LockedQueue<T>::NOT_STRICT_SHARDING);
+    EnsureShardExists(shard);
+    if (index_[shard].tasks.size()) {
```
matthewtff: Use `!empty()` here.

src/daemon/emitter_test.cc
```diff
@@ -456,6 +456,246 @@ TEST_F(EmitterTest, ConfigurationUpdateFromCoordinator) {
...
+        const ui32 expected_shard = Emitter::CalculateShard(
+            cache::string::HandledHash(handled_hash), old_total_shards);
+
+        EXPECT_EQ(EndPointString(remote_host_name, expected_shard),
+                  end_point->Print());
+
+        // Ensure that shard is greater than new total shards we're sending in
+        // new configuration, to make sure the second task will be
+        // redistributed.
+        EXPECT_LT(new_total_shards, expected_shard);
```
matthewtff: Change to `ASSERT_LT` with appropriate text.

src/daemon/emitter_test.cc
```diff
@@ -456,6 +456,246 @@ TEST_F(EmitterTest, ConfigurationUpdateFromCoordinator) {
...
+      });
+
+      connection->CallOnRead([&](net::Connection::Message* message) {
+        // Do not send any code to prevent task result appearing in cache to
+        // make sure next task with same handled source code goes to remote.
+
+        // Make sure coordinator thread starts new pool, stops
+        // current and redistributes tasks.
+        // Cant wait here as coordinator should join this thread.
+        std::this_thread::sleep_for(std::chrono::milliseconds(100));
```
matthewtff: Change to `Seconds(1)`

PR #112, opened 2017-06-27
Update libc++.

matthewtff:


matthewtff:
Failed test looks like a flake. I've build it on my linux host and it doesn't fail.

src/third_party/libcxx/BUILD.gn
```diff
@@ -1,3 +1,8 @@
+# Used cmake args:
+# -DLIBCXX_INCLUDE_BENCHMARKS=OFF
```
abyss7: You should list all the non-CMAKE args here - in case their default values change in future versions.

src/third_party/libcxx/BUILD.gn
```diff
@@ -142,38 +235,163 @@ shared_library("c++") {
...
+    "exported/include/cfloat",
+    "exported/include/chrono",
+    "exported/include/cinttypes",
+    "exported/include/ciso646",
+    "exported/include/climits",
+    "exported/include/clocale",
+    "exported/include/cmath",
+    "exported/include/codecvt",
+    "exported/include/complex",
+    "exported/include/complex.h",
```
abyss7: Are you sure that `.h` files are the part of public interface? Are they copied to public `include` folder when you install libcxx with CMake?

abyss7:
You should provide info about commit used to update libcxx.

PR #113, opened 2017-06-30
Move file operations to <filesystem> where possible

matthewtff:


src/base/file/file_posix.cc
```diff
@@ -69,29 +96,19 @@ bool File::IsExecutable(const String& path, String* error) {
...
   DCHECK(IsValid());
 
-  struct stat buffer;
-  if (fstat(native(), &buffer)) {
-    GetLastError(error);
-    return 0;
-  }
-
-  return buffer.st_size;
+  return File::Size(path_, error);
```
matthewtff: This one may be a bad idea =/

src/base/file_utils.cc
```diff
@@ -0,0 +1,111 @@
...
+#include STL(chrono)
+#include STL_EXPERIMENTAL(filesystem)
+#include STL(map)
+#include STL(regex)
+#include STL(system_error)
+
+namespace dist_clang {
+namespace base {
+
+void WalkDirectory(
```
abyss7: This is a hard part - I'll review this function later.

src/base/file_utils.cc
```diff
@@ -0,0 +1,111 @@
...
+  WalkDirectory(path, [&result](const Path&, ui64, ui64 size) {
+    result += size;
+  }, error);
+
+  return result;
+}
+
+bool RemoveEmptyDirectory(const Path& path) {
+  std::error_code ec;
+  return std::experimental::filesystem::remove(path, ec);
```
abyss7: We can accidentally pass the file and remove it!

src/base/file_utils.cc
```diff
@@ -0,0 +1,111 @@
...
+    GetLastError(error);
+    return false;
+  }
+  return true;
+#endif
+}
+
+bool CreateDirectory(const Path& path, String* error) {
+  std::error_code ec;
+  std::experimental::filesystem::create_directories(path, ec);
```
abyss7: It sets 0777 perms by default, while we want 0755

src/base/file_utils_test.cc
```diff
@@ -126,15 +66,13 @@ TEST(FileUtilsTest, TempFile) {
 TEST(FileUtilsTest, CreateDirectory) {
   String error;
   const base::TemporaryDir temp_dir;
-  const String temp = String(temp_dir) + "/1/2/3";
+  const Path& temp = temp_dir.GetPath() / "1" / "2" / "3";
 
-  ASSERT_TRUE(CreateDirectory(temp, &error)) << error;
+  ASSERT_TRUE(CreateDirectory(temp.string(), &error)) << error;
```
abyss7: Why `.string()`?

src/base/file/file_posix.cc
```diff
@@ -15,16 +19,41 @@
...
+  return base::AppendExtension(path, ".tmp");
+}
+
+}  // anonymous namespace
+
 namespace base {
 
-File::File(const String& path) : Data(open(path.c_str(), O_RDONLY)) {
+File::File(const Path& path)
+    : Data(open(path.c_str(), O_RDONLY | O_CLOEXEC)), path_(path) {
```
matthewtff: Why wasn't it here before?)

src/base/file/file_posix.cc
```diff
@@ -45,18 +74,16 @@ File::File(const String& path) : Data(open(path.c_str(), O_RDONLY)) {
...
-  if (stat(path.c_str(), &buffer)) {
-    GetLastError(error);
+bool File::IsFile(const Path& path, String* error) {
+  std::experimental::filesystem::file_status status;
+  if (!GetStatus(path, &status, error)) {
     return false;
   }
-
-  return (buffer.st_mode & S_IFDIR) == 0;
+  return std::experimental::filesystem::is_regular_file(status);
```
matthewtff: Ok, I'll change it to `!is_directory`

src/base/file/file_posix.cc
```diff
@@ -69,29 +96,19 @@ bool File::IsExecutable(const String& path, String* error) {
...
-
-  if (!IsFile(path, error)) {
+bool File::Exists(const Path& path, String* error) {
+  std::experimental::filesystem::file_status status;
+  if (!GetStatus(path, &status, error)) {
     return false;
   }
-
-  return true;
+  return std::experimental::filesystem::exists(status) &&
```
matthewtff: We anyway need a `status()` inside `IsFile` function, so looks like not a big deal.
Regression only in case when `access` return non-zero which should be rare.

src/base/file/file_posix.cc
```diff
@@ -284,51 +304,29 @@ bool File::Copy(const String& src_path, const String& dst_path, String* error) {
...
         // We need write-access even on object files after introduction of the
         // "split-dwarf" option, see
         // https://sourceware.org/bugzilla/show_bug.cgi?id=971
         const auto mode = mode_t(S_IRUSR | S_IWUSR | S_IRGRP | S_IROTH);
         const auto flags = O_WRONLY | O_CREAT | O_TRUNC | O_CLOEXEC;
-        const String tmp_path = path + ".tmp";
+        const Path tmp_path = AddTempEnding(path);
         return open(tmp_path.c_str(), flags, mode);
       }()),
+      path_(AddTempEnding(path)),
```
matthewtff: Yeah, I'll remove this logic.

src/base/file/file_test.cc
```diff
@@ -5,21 +5,24 @@
...
+#include STL(system_error)
 
 namespace dist_clang {
 namespace base {
 
 TEST(FileTest, Read) {
   const auto expected_content = "All your base are belong to us"_l;
   const base::TemporaryDir temp_dir;
-  const String file_path = String(temp_dir) + "/file";
+  const Path temp_dir_path = temp_dir.GetPath();
```
matthewtff: As `tmp_dir.GetPath()` looks more explicit than `operator Path()`.
But that's matter of taste, so let get the operator way...

src/base/file_utils.h
```diff
@@ -1,43 +1,24 @@
...
+
+ui64 CalculateDirectorySize(const Path& path, String* error = nullptr);
+
+bool RemoveEmptyDirectory(const Path& path);
+
+bool ChangeOwner(const Path& path, ui32 uid, String* error = nullptr);
+
+bool CreateDirectory(const Path& path, String* error = nullptr);
+
+Path AppendExtension(Path path, const char* extension);
```
matthewtff: I believe it's intentionally to make it difficult to misuse with std::string.

src/base/c_utils.h
```diff
@@ -52,5 +59,19 @@ inline void GetLastError(String* error) {
   }
 }
 
+inline bool SetPermissions(const Path& path, const Perms permissions,
```
matthewtff: It should fit the machine word, but I'll make it a const ref no difference for me.

src/base/file/file_posix.cc
```diff
@@ -235,17 +253,19 @@ bool File::Hash(const String& path, Immutable* output,
 }
 
 // static
-bool File::Delete(const String& path, String* error) {
-  if (unlink(path.c_str()) == -1) {
-    GetLastError(error);
+bool File::Delete(const Path& path, String* error) {
+  std::error_code ec;
+  if (!std::experimental::filesystem::remove(path, ec)) {
```
matthewtff: I'll switch this to `std::remove` from `cstdio` then.

src/base/file_utils.cc
matthewtff: Is it OK to add a `std::fs::is_directory` check, or you would like to leave previous code for performance reasons?

src/base/file_utils.cc
matthewtff: Looks like there's no way to implement this using <filesystem> :(
I'll revert this change, thanks.

src/base/const_string.h
```diff
@@ -16,6 +16,7 @@ namespace base {
 
 class Literal {
  public:
+  inline const char* c_str() const { return str_; }
```
matthewtff: In `ReplaceTildeInPath` from file_cache.cc.
`GetHomeDir()` returns a `Literal` thats why I should use `c_str()` to create a `Path` form it.

matthewtff:
Should I close this one now?

abyss7:
No, I want to use this PR as a reference for a further work.

PR #114, opened 2017-07-24
Added cache tasks to absorber

matthewtff:
Absorber has a limited queue of incoming tasks and if new task doesn't
fit the queue, the task is rejected. New cache tasks change this
behavior to first check if there's a result for task in cache. This
should increase of cache use while absorbers are overloaded.

### DIFF
--- a/src/daemon/absorber.cc
+++ b/src/daemon/absorber.cc
@@ -15,13 +15,24 @@ using namespace std::placeholders;
 namespace dist_clang {
 namespace daemon {
 
+namespace {
+const char kOverloadedErrorText[] = "Tasks queue reached limit";
+}  // anonymous namespace
+
 Absorber::Absorber(const Configuration& conf) : CompilationDaemon(conf) {
   using Worker = base::WorkerPool::SimpleWorker;
   CHECK(conf.has_absorber() && !conf.absorber().local().disabled());
 
   workers_ = std::make_unique<base::WorkerPool>();
+  cache_tasks_ = std::make_unique<Queue>();
   tasks_ = std::make_unique<Queue>(conf.pool_capacity());
 
+  if (conf.has_cache() && !conf.cache().disabled()) {
+    Worker worker = std::bind(&Absorber::DoCheckCache, this, _1);
+    const auto threads_number = conf.cache().has_threads()
+        ? conf.cache().threads() : std::thread::hardware_concurrency();
+    workers_->AddWorker("Cache Worker"_l, worker, threads_number);
+  }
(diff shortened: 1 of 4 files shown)


src/daemon/absorber.cc
```diff
@@ -15,13 +15,24 @@ using namespace std::placeholders;
 namespace dist_clang {
 namespace daemon {
 
+namespace {
+const char kOverloadedErrorText[] = "Tasks queue reached limit";
+}  // anonymous namespace
```
abyss7: `//  namespace`

src/daemon/absorber.cc
```diff
@@ -71,15 +85,19 @@ bool Absorber::HandleNewMessage(net::ConnectionPtr connection,
     Message execute(message->ReleaseExtension(proto::Remote::extension));
     DCHECK(!execute->flags().compiler().has_path());
     if (execute->has_source()) {
-      if (tasks_->Push(Task{connection, std::move(execute)})) {
-        return true;
+      Task task{connection, std::move(execute), HandledHash()};
+      if (conf->has_cache() && !conf->cache().disabled()) {
+        cache_tasks_->Push(std::move(task));
       } else {
```
abyss7: ` } else if (…) {` ?

src/daemon/absorber.cc
```diff
@@ -71,15 +85,19 @@ bool Absorber::HandleNewMessage(net::ConnectionPtr connection,
     Message execute(message->ReleaseExtension(proto::Remote::extension));
     DCHECK(!execute->flags().compiler().has_path());
     if (execute->has_source()) {
-      if (tasks_->Push(Task{connection, std::move(execute)})) {
-        return true;
+      Task task{connection, std::move(execute), HandledHash()};
```
abyss7: Put some comment that we plan in the future to take handled hash from incoming message.

src/daemon/absorber.cc
```diff
@@ -124,40 +155,37 @@ bool Absorber::PrepareExtraFilesForCompiler(
...
 
   while (!pool.IsShuttingDown()) {
-    Optional&& task = tasks_->Pop();
+    Optional&& task = cache_tasks_->Pop();
     if (!task) {
       break;
     }
 
-    if (task->first->IsClosed()) {
+    // Create a copy of pointer to connection here to be able to send responce
```
abyss7: I commited the change to `LockedQueue::Push()` so it shouldn't corrupt objects in case of failure. No need for local copy of `connection`.

PR #116, opened 2017-08-15
Added support for rewrite-includes flag while preprocessing

matthewtff:


src/base/base.proto
```diff
@@ -49,6 +49,9 @@ message Flags {
   repeated string included_files = 12;
   // PCH or PTH files, included implicitly through command line arguments.
 
+  optional bool rewrite_includes = 13;
+  // Indicates if preprocessing should/was made using frewrite-includes flag.
```
abyss7: '-frewrite-includes' flag ?

src/client/clang.cc
```diff
@@ -20,7 +20,8 @@ namespace client {
 bool DoMain(int argc, const char* const argv[], Immutable socket_path,
             Immutable clang_path, Immutable version, ui32 connect_timeout_secs,
             ui32 read_timeout_secs, ui32 send_timeout_secs, ui32 read_min_bytes,
-            const HashMap<String, String>& plugins, bool disabled) {
+            const HashMap<String, String>& plugins, bool disabled,
+            const bool rewrite_includes) {
```
abyss7: Why `const` for function argument here? At least it's not a common style in other parts of code.

src/base/base.proto
matthewtff: Flag is passed without dash to cc1 or driver, so I intentionally skipped it every time it is mentioned. I'll take it in quotes, no problem here.

src/client/clang_command.cc
```diff
@@ -13,6 +13,11 @@
 namespace dist_clang {
 namespace client {
 
+// Flags to be ignored that are known to somehow break compilation.
+static const std::array<clang::driver::options::ID, 1> kIngoredFlags = {
```
abyss7: No need for static inside '.cc' - use anonymous namespace

src/client/clang_command.cc
```diff
@@ -13,6 +13,11 @@
...
+  {clang::driver::options::OPT_frewrite_includes},
```
abyss7: Why not to add -frewrite-macros too?

src/client/clang_command.cc
```diff
@@ -10,6 +10,16 @@
 
 #include <base/using_log.h>
 
+namespace {
+
+// Flags to be ignored that are known to somehow break compilation.
+const std::array<clang::driver::options::ID, 2> kIngoredFlags = {{
```
abyss7: Fix typo: `IgnoredFlags`

src/client/clang_command.cc
```diff
@@ -155,10 +171,23 @@ bool ClangCommand::FillFlags(base::proto::Flags* flags,
 
     // By default all other flags are cacheable.
     else {
-      arg->render(arg_list_, other_list);
+      const bool ignored = std::any_of(kIngoredFlags.begin(),
```
abyss7: Add `FIXME` about potential O(n*m) complexity problem.

src/client/command_test.cc
```diff
@@ -190,8 +191,74 @@ TEST(CommandTest, FillFlags) {
   EXPECT_EQ("-cc1", *flags.other().begin());
   EXPECT_EQ(1, flags.compiler().plugins_size());
   EXPECT_EQ(plugin_name, flags.compiler().plugins(0).name());
+
+  // Check that macroses go to non-cached list.
+  EXPECT_NE(flags.non_cached().end(),
+            std::find(flags.non_cached().begin(), flags.non_cached().end(),
+                      "MACRO"));
```
abyss7: Why do you search for "MACRO" and not "-DMACRO" ?

src/client/command_test.cc
```diff
@@ -190,8 +191,74 @@ TEST(CommandTest, FillFlags) {
...
+
+  // Make sure frewrite-includes removed if not used explicitly.
+  EXPECT_EQ(flags.other().end(),
+            std::find(flags.other().begin(), flags.other().end(),
+                      "frewrite-includes"));
```
abyss7: Why do you check for "frewrite-includes" and not "-frewrite-includes" ?

src/client/command_test.cc
matthewtff: Tehically `-D` and `MACRO` are different flags. You can see an example here: http://rextester.com/LFEJ65675 (expand compiler flags).
So I believe it's enough for test to test just for `MACRO` existence in flags.
If you like, I could add two more expects: that there is also a `-D` flag, and that it precedes the `MACRO` flag.

src/client/command_test.cc
matthewtff: As it is really a `frewrite-includes` flag, without leading dash.

src/daemon/compilation_daemon.cc
```diff
@@ -44,6 +44,9 @@ inline CommandLine CommandLineForSimpleCache(const base::proto::Flags& flags) {
     command_line += " " + base::JoinString<' '>(flags.cc_only().begin(),
                                                 flags.cc_only().end());
   }
+  if (flags.has_rewrite_includes()) {
+    command_line += " frewrite-includes"_l;
```
abyss7: We render this flag via `other` list - why to append it explicitly here?

src/client/clang_command.cc
```diff
@@ -155,10 +171,27 @@ bool ClangCommand::FillFlags(base::proto::Flags* flags,
...
+        arg->render(arg_list_, other_list);
+      }
     }
   }
 
+  if (rewrite_includes) {
+    const auto& option = opts_->getOption(OPT_frewrite_includes);
+    String option_name = tmp_list.MakeArgString(option.getRenderName());
+    option_name.insert(0, 1, '-');
+    other_list.push_back(option_name.c_str());
```
abyss7: Does it make `strcpy` inside?

matthewtff: Yes, it does.

matthewtff: No, nice catch

